PR #1, opened 2022-07-12
Initial testing framework for substrait consumer testing.

richtia:
Currently only has tests with acero as a consumer

Tests take a file, sql query, and substrait query plan as input.
Run the substrait query plan on Acero
Run the SQL query against DuckDB
Compare results.

test_acero_tpch.py:
Test class file

basetest.py:
Base test to be subclassed by other tests.  Includes methods for basic verification.

verification.py:
Implementation of verification methods.

common.py:
Common functions that will be used by tests.

parametrization.py:
Wrapper around pytest's default test case parametrization decorator to make test cases more readable.

consumer.py:
Abstract base class for implementing additional consumers

acero_consumer.py:
implementation of acero consumer class


rok:
@richtia would it make sense to have tpch queries (tests/integration/queries/tpch/query_x.py) stored in json or yaml?

richtia:
> 

Yeah, if anyone else thinks they could use the substrait plan in these files, I was thinking later on I could just have the substrait plan as a separate json file that the .py testcase files imports.

vibhatha:
@richtia would it be better to store the parquet files in a separate repo and may be use a submodule to represent the usage?

jacques-n:
The big concern I have with the existing suggested approach is that I've seen several conversations that suggest DuckDB plans are not consistent with the spec in at least a few ways. (For example, there was some recent mention to an incorrect handling of project and emit within a email list thread last week.)

I'd like to suggest that we start first with a human validated plans get those working across systems before adding something that may or not be actually correct like the above. (For reference, I think that the way this was done for the validator repo was to generate stuff using an existing tool and then hand check/fix the plans to match the specification.)

richtia:
> The big concern I have with the existing suggested approach is that I've seen several conversations that suggest DuckDB plans are not consistent with the spec in at least a few ways. (For example, there was some recent mention to an incorrect handling of project and emit within a email list thread last week.)
> 
> I'd like to suggest that we start first with human validated plans get those working across systems before adding something that may or not be actually correct like the above. (For reference, I think that the way this was done for the validator repo was to generate stuff using an existing tool and then hand check/fix the plans to match the specification.)

Just to clarify, verification is done here by calculating the results of running the SQL query against DuckDB, not the substrait plan against DuckDB.  The substrait query plan is only run against Acero, we just need something to use as a source of truth to verify those results against.

Doing some functional testing with smaller human validated plans, would also be a good starting point.

CC: Ian/Weston

westonpace:
> The big concern I have with the existing suggested approach is that I've seen several conversations that suggest DuckDB plans are not consistent with the spec in at least a few ways. (For example, there was some recent mention to an incorrect handling of project and emit within a email list thread last week.)
> 
> I'd like to suggest that we start first with human validated plans get those working across systems before adding something that may or not be actually correct like the above. (For reference, I think that the way this was done for the validator repo was to generate stuff using an existing tool and then hand check/fix the plans to match the specification.)

I think both approaches are useful.  The TPC-H "SQL" has some pretty well known and standard answers.  However, there is always room for error in how that SQL is written or in how that SQL translated to Substrait.  I think that making sure engines agree on larger / random queries has value in a sort of exploratory, exposing differences that were previously unknown way.

On the other hand, there is likely to be a need for smaller queries that can serve to "codify" parts of the spec in the same way that the validator does.

jacques-n (commented):
I misunderstood what you're doing here. My question now would be where did the plans come from in these tests. My main caution is having plans in a canonical repo that we haven't validated. That increases the likelihood of divergence of the spec.

richtia:
> I misunderstood what you're doing here. My question now would be where did the plans come from in these tests. My main caution is having plans in a canonical repo that we haven't validated. That increases the likelihood of divergence of the spec.

I generated these plans by running the SQL through Isthmus.  I wanted to have the plans in plain text as opposed to the byte representation so I could see what they contain.

tests/integration/queries/tpch/query_1.py
```diff
@@ -0,0 +1,844 @@
+import pytest
+
+TESTCASE = [
```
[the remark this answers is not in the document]
richtia: I originally put these as python files so that I could use some pytest decorators for marking tests, i.e., skipping one that fails due to a known bug.

I can separate the substrait plan portion into it's own json file if that makes things easier.

jacques-n:
> I generated these plans by running the SQL through Isthmus. I wanted to have the plans in plain text as opposed to the byte representation so I could see what they contain.

Got it. Do they validate against the validator? There were a couple of recent fixes in Isthmus to address invalid plan attributes.

richtia:
> > I generated these plans by running the SQL through Isthmus. I wanted to have the plans in plain text as opposed to the byte representation so I could see what they contain.
> 
> Got it. Do they validate against the validator? There were a couple of recent fixes in Isthmus to address invalid plan attributes.

I haven't tried validating them against the validator yet...but I was planning to add that as part of these test suites as well.  @westonpace had mentioned this to me.  I was planning on doing this in another PR, but I can add it as part of this one instead.

jacques-n:
> > > I generated these plans by running the SQL through Isthmus. I wanted to have the plans in plain text as opposed to the byte representation so I could see what they contain.
> > 
> > 
> > Got it. Do they validate against the validator? There were a couple of recent fixes in Isthmus to address invalid plan attributes.
> 
> I haven't tried validating them against the validator yet...but I was planning to add that as part of these test suites as well. @westonpace had mentioned this to me. I was planning on doing this in another PR, but I can add it as part of this one instead.

To start, I'd just want to know that they were validated. A subsequent patch can prove that. (Maybe just start with doing it by hand.)

richtia:
> > > > I generated these plans by running the SQL through Isthmus. I wanted to have the plans in plain text as opposed to the byte representation so I could see what they contain.
> > > 
> > > 
> > > Got it. Do they validate against the validator? There were a couple of recent fixes in Isthmus to address invalid plan attributes.
> > 
> > 
> > I haven't tried validating them against the validator yet...but I was planning to add that as part of these test suites as well. @westonpace had mentioned this to me. I was planning on doing this in another PR, but I can add it as part of this one instead.
> 
> To start, I'd just want to know that they were validated. A subsequent patch can prove that. (Maybe just start with doing it by hand.)

So here are some issues I'm seeing with running the query 1 substrait plan against the validator:
1.  It doesn't seem to like the "parquet" field name part of the LocalFiles ReadRel
```
Message type "substrait.ReadRel.LocalFiles.FileOrFiles" has no field named "parquet" at "Plan.relations[0].root.input.sort.input.aggregate.input.project.input.filter.input.read.local_files.items[0]"
```
2.  If I change it back to namedtable instead of the LocalFiles, I run into a ValueError
```
ValueError: Error at plan.relations[0].rel_type<root>: mismatched field name associations: received 1 too few field name(s) (code 4003)
```

cc: @jvanstraten

jvanstraten:
Sorry, the change in how file formats are represented is a breaking change that was introduced in Substrait after I got blocked on maintaining the validator due to lack of reviews and Substrait's review policy :( See also https://github.com/substrait-io/substrait-validator/issues/19

jvanstraten:
Without even looking at the plans, I'm willing to bet both are due to the grouping set index column not existing according to Isthmus. I've double-checked since the sync call that the spec doesn't say that column's existence is conditioned on there being more than one grouping set:

> The list of distinct columns from each grouping set (ordered by their first appearance) followed by the list of measures in declaration order, followed by an integer describing the associated particular grouping set the value is derived from.

The third is just a warning and can be ignored; you'll get not-yet-implemented warnings from the validator anyway. This particular warning occurs because the validator wants valid, resolvable URLs (while Substrait allows any URI) unless you tell it how to resolve them via a callback or overrides. AFAIK there's no real consensus yet about what the extension URIs should look like, so the validator is still pretty naive there.

richtia:
> Without even looking at the plans, I'm willing to bet both are due to the grouping set index column not existing according to Isthmus. I've double-checked since the sync call that the spec doesn't say that column's existence is conditioned on there being more than one grouping set:
> 
> > The list of distinct columns from each grouping set (ordered by their first appearance) followed by the list of measures in declaration order, followed by an integer describing the associated particular grouping set the value is derived from.
> 
> The third is just a warning and can be ignored; you'll get not-yet-implemented warnings from the validator anyway. This particular warning occurs because the validator wants valid, resolvable URLs (while Substrait allows any URI) unless you tell it how to resolve them via a callback or overrides. AFAIK there's no real consensus yet about what the extension URIs should look like, so the validator is still pretty naive there.

Thanks for taking a look.  What changes are needed to get this to pass?  Do I need to modify the substrait plan somewhere or does this need to be fixed with a change on the substrait/Isthmus side?

jvanstraten:
You can use the "check not invalid" version of the check function *or* use the more advanced interface with which you can pass options to supress the warnings you're getting (for now) to stop it from failing on warnings. The aggregates should be fixed in Isthmus; whether or not you want to fix the plans manually for the time being is up to you. You just need to add an emit clause to the aggregation that lets all columns except the grouping set index column through. You can use the HTML output of the validator to see how many columns that is.

tests/common.py
```diff
@@ -0,0 +1,124 @@
...
+        [os.path.join(data_dir, dataset) for dataset in file_names]
+    )
+
+    return full_paths_list
+
+
+def get_substrait_plan(filename: str) -> str:
+    realpath_directory = os.path.realpath(
+        os.path.join(os.getcwd(), os.path.dirname(__file__)))
+    plan_dir = os.path.join(realpath_directory, 'integration/queries/tpch_substrait_plans')
```
gforsyth, line 39: This `realpath_directory` shows up in a couple of places -- since it will be the same for this file, you could define it at the top as a module-level variable and re-use it.

tests/common.py
```diff
@@ -0,0 +1,124 @@
...
+    plan_path = os.path.join(plan_dir, filename)
```
gforsyth, line 40:
```suggestion
    plan_path = os.path.join(realpath_directory, 'integration/queries/tpch_substrait_plans', filename)
```

You can pass arbitrarily many paths to `os.path.join` to construct longer paths -- also note that `'integration/queries/tpch_substrait_plans/` is relying on `/` being a valid path separator, so this won't work on Windows (that may not be a concern)

tests/common.py
```diff
@@ -0,0 +1,124 @@
...
+
+    f = open(plan_path, 'r')
+    plan = json.loads(f.read())
+
+    return json.dumps(plan)
```
gforsyth, line 45:
```suggestion
    with open(plan_path, "r") as f:
        return f.read()
```

I may be misunderstanding what's happening here -- but right now we load from a string into a json object, then dump it out as a string.  If the idea is to validate the JSON, that should be called out in a separate function.  Otherwise, this could just be returning the contents of the file as a string.

tests/common.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class SubstraitUtils(object):
```
gforsyth, line 48:
```suggestion
class SubstraitUtils:
```

tests/common.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self, name="SubstraitUtils"):
+        """
+
+        """
+        self.logger = LOGGER
+        self.logger.info(name)
+
+    @staticmethod
+    def arrow_sort_tb_values(table: type[Table],
+                             sortby: list[str]) -> type[Table]:
```
gforsyth, line 61:
```suggestion
    def arrow_sort_tb_values(table: pa.Table,
                             sortby: list[str]) -> pa.Table:
```

tests/common.py
```diff
@@ -0,0 +1,124 @@
+import logging
+import json
+import os
+
+import pyarrow as pa
+from pyarrow.lib import tobytes
+from pyarrow.lib import Table
```
gforsyth, line 7:
```suggestion
```
I would make use of `pa` here, otherwise `Table` seems pretty generic.  And there's a `pa.Table` available that points to `pa.lib.Table`

tests/common.py
```diff
@@ -0,0 +1,124 @@
...
+        # Get full path for all datasets used in the query
+        parquet_file_paths = get_full_path(file_names)
+
+        # Replace the filename placeholder in the substrait query plan with
+        # the proper parquet data file paths.
+        for count, file_path in enumerate(parquet_file_paths):
+            substrait_query = substrait_query.replace(
+                f"FILENAME_PLACEHOLDER_{count}", file_path)
+
+        return tobytes(substrait_query)
```
gforsyth, line 124:
```suggestion
        return pa.lib.tobytes(substrait_query)
```
(since I removed the import above)

tests/consumers/acero_consumer.py
```diff
@@ -0,0 +1,52 @@
...
+
+from .consumer import AbstractConsumer
+
+try:
+    import pyarrow.substrait as substrait
+except ImportError:
+    substrait = None
+
+
+class AceroConsumer(AbstractConsumer):
```
gforsyth, line 12: We chatted offline -- I'm not sure I see the value in the abstract base class at this stage.

tests/consumers/acero_consumer.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Implementation of the acero substrait consumer Class for testing
+    """
+    def __init__(self):
+        pass
+
+    @staticmethod
+    def run_substrait_query(substrait_query: str) -> type[Table]:
```
gforsyth, line 20:
```suggestion
    def run_substrait_query(substrait_query: str) -> pa.Table:
```

tests/consumers/acero_consumer.py
```diff
@@ -0,0 +1,52 @@
...
+        Run the given substrait query and return the result
+
+        Args:
+            substrait_query: A json formatted string representing the
+            substrait query plan
+
+        Returns:
+            A pyarrow table resulting from running the substrait query plan.
+        """
+        buf = pa._substrait._parse_json_plan(substrait_query)
```
gforsyth, line 31: This looks like a private module -- is there a public version of this you can hit?

gforsyth: And if there isn't, can `pyarrow` add one?  You shouldn't be required to rely on a private API to do this (to my mind)

tests/consumers/acero_consumer.py
```diff
@@ -0,0 +1,52 @@
...
+        reader = substrait.run_query(buf)
+        result = reader.read_all()
+
+        return result
+
+    @staticmethod
+    def write_to_arrow_binary_file(table: type[Table], filepath: str) -> None:
```
gforsyth, line 38:
```suggestion
    def write_to_arrow_binary_file(table: pa.Table, filepath: Path | str) -> None:
```

You'll have to add a `from __future__ import annotations` to the top of this file for the `|` union operator to work, but it's a good idea to allow both `pathlib.Path` objects and also `str` for file locations

tests/consumers/consumer.py
```diff
@@ -0,0 +1,13 @@
...
+
+from pyarrow.lib import Table
+
+
+class AbstractConsumer(ABC):
+    """
+    Abstract class for representing a substrait consumer
+    """
+    @abstractmethod
+    def run_substrait_query(self, substrait_query: str) -> type[Table]:
```
gforsyth, line 12:
```suggestion
    def run_substrait_query(self, substrait_query: str) -> pa.Table:
```

tests/integration/queries/tpch/query_1.py
```diff
@@ -0,0 +1,34 @@
...
+            from
+              '{}'
+            where
+              l_shipdate <= date '1998-12-01' - interval '120' day (3)
+            group by
+              l_returnflag,
+              l_linestatus
+            order by
+              l_returnflag,
+              l_linestatus
```
gforsyth, line 30: similar comment as @jacques-n above -- I would have these live in separate `sql` files, makes these files cleaner and also provides an easy reference.

tests/verification.py
```diff
@@ -0,0 +1,75 @@
...
+
+# add the handler to the logger
+logging.getLogger('').addHandler(LOG_HANDLER)
+
+
+class Verifier(object):
+    """
+
+    """
+    def __init__(self, name="Verifier"):
```
gforsyth, line 23:
```suggestion
    def __init__(self, name: str = "Verifier") -> None:
```

tests/verification.py
```diff
@@ -0,0 +1,75 @@
...
+
+# create formatter and add it to the handler
+FORMATTER = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+LOG_HANDLER.setFormatter(FORMATTER)
+
+# add the handler to the logger
+logging.getLogger('').addHandler(LOG_HANDLER)
+
+
+class Verifier(object):
```
gforsyth, line 19:
```suggestion
class Verifier:
```

tests/verification.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+
+    """
+    def __init__(self, name="Verifier"):
+        """
+
+        """
```
gforsyth, line 26:
```suggestion
```

tests/verification.py
```diff
@@ -0,0 +1,75 @@
...
+        Returns:
+            None
+
+        Raises:
+            AssertionError: An error occurred.
+
+        """
+        raise AssertionError(f"TEST FAILURE: {message}")
+
+    def verify_equals(self, actual: object, expected: object,
```
gforsyth, line 46:
```suggestion
    def verify_equals(self, actual: Any, expected: Any,
```

tests/verification.py
```diff
@@ -0,0 +1,75 @@
...
+        msg = [f"Verifying equals: {actual} == {expected}."]
+        msg = [message] if message else msg
+
+        if type(actual) != type(expected):
+            msg = (
+                f"Object types are not the same. \nActual type: "
+                f"{type(actual)}\nExpected type: {type(expected)}"
+            )
+            self.fail(msg)
+        elif actual != expected:
```
gforsyth, line 72: is this `Verifier` going to be subclassed?  For a lot of objects I imagine this code will be comparing, the python `==` or `!=` aren't going to do the expected thing.

gforsyth (commented):
A few questions, a few type-hint fixes, and some nitpicks around doc-strings.

tests/consumers/acero_consumer.py
richtia: @vibhatha is there a public version of this?  I couldn't find one.

vibhatha: @richtia there was one I wrote for the initial PR, but we sort of made it private. The reason was we never wanted to expose it as a public feature at the time of the public API. But this doesn't restrict us from adding one as a utility. 
cc @westonpace shall we create a util API for Substrait in Python including this feature. I can create a quick PR for this if this is okay.

tests/common.py
vibhatha: Agree, in the public API we tend to use `from pyarrow import Table`

tests/verification.py
```diff
@@ -0,0 +1,76 @@
...
+                Message to be logged.
+
+        Raises:
+            AssertionError:
+                An error occurred.
+
+        """
+        raise AssertionError(f"TEST FAILURE: {message}")
+
+    def verify_equals(self, actual: Any, expected: Any,
```
vibhatha, line 43: suggestion: It would better to write [`__eq__`](https://docs.python.org/3/library/operator.html#operator.eq) operator here rather than writing a regular method. Then the objects extended for this capability would be easy to use and could extend specific features to be compared to the sub-classes if required.

tests/basetest.py
```diff
@@ -0,0 +1,57 @@
...
+        Log the message and raise an AssertionError
+
+        Parameters:
+            message:
+                message to be logged.
+        """
+        cls.verifier.fail(message)
+
+    @classmethod
+    def verify_equals(cls, actual: Any, expected: Any,
```
vibhatha, line 42: should we use `__eq__` here? cc @gforsyth

tests/common.py
```diff
@@ -0,0 +1,121 @@
...
+
+REALPATH_DIRECTORY = os.path.realpath(os.path.join(os.getcwd(), os.path.dirname(__file__)))
+
+
+def get_full_path(file_names: Iterable[str]) -> list[str]:
+    data_dir = os.path.join(REALPATH_DIRECTORY, 'data/tpch_parquet')
+    full_paths_list = (
+        [os.path.join(data_dir, dataset) for dataset in file_names]
+    )
+
```
vibhatha, line 31: we could get rid of this spacing?

tests/common.py
```diff
@@ -0,0 +1,121 @@
...
+        return f.read()
+
+
+class SubstraitUtils:
+    """
+    Common utility for substrait integration tests.
+    """
+    def __init__(self, name="SubstraitUtils"):
+        """
+
```
vibhatha, line 49: missing documentation?

tests/consumers/acero_consumer.py
```diff
@@ -0,0 +1,52 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+import pyarrow as pa
+
+try:
+    import pyarrow.substrait as substrait
+except ImportError:
+    substrait = None
```
vibhatha, line 10: I assume PyArrow built with Substrait is a requirement for this package or is it optional? 
If it is not optional we should probably have this just as `import pyarrow.substrait as substrat` or `import pyarrow.substrait as sbt`

tests/consumers/acero_consumer.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class AceroConsumer:
+    """
+    Implementation of the acero substrait consumer Class for testing
+    """
+    def __init__(self):
+        pass
```
vibhatha, line 18: why just pass?

tests/integration/queries/tpch/query_3.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+from ....common import get_substrait_plan
```
vibhatha, line 2: Hmm.... may be once we are cleared with the functionality and addressing majority of the concerns. It would be better to package this. Include a `setup.py` and other required components. Then we can release the functional consumer test suite nightly and a stable release in matched with Substrait, Arrow, Duckdb, ... releases. 


cc @gforsyth @jacques-n @rok @jvanstraten

tests/consumers/acero_consumer.py
richtia: Removed the abstract base class

tests/consumers/acero_consumer.py
richtia: I actually don't even need this.  I'll remove it.

westonpace (commented):
* Can you add a README.MD at the top level describing the purpose of the repository, explain the general technologies used (e.g. pytest), and some basic instructions how to use it?
* Can you add some kind of requirements.txt file or something similar that helps communicate what python dependencies are required to run these tests?
* The test output is very noisy.  I was playing with this earlier today and failed all five tests (as expected) but had a hard time seeing the test names that failed, for example.  The log output is likely useful for debugging but a more succinct test report would be nicer by default.
* Can you add a separate readme file to the data/tpch_parquet directory explaining how these files were generated.  Also, these files end with _1 but I don't think they are scale factor 1 (are these the 0.1 files we generated?)

vibhatha:
> * Can you add a README.MD at the top level describing the purpose of the repository, explain the general technologies used (e.g. pytest), and some basic instructions how to use it?
> * Can you add some kind of requirements.txt file or something similar that helps communicate what python dependencies are required to run these tests?
> * The test output is very noisy.  I was playing with this earlier today and failed all five tests (as expected) but had a hard time seeing the test names that failed, for example.  The log output is likely useful for debugging but a more succinct test report would be nicer by default.
> * Can you add a separate readme file to the data/tpch_parquet directory explaining how these files were generated.  Also, these files end with _1 but I don't think they are scale factor 1 (are these the 0.1 files we generated?)

@westonpace how about packaging this as a regular python package and include some CIs if possible. Down the road it would be important? WDYT?

richtia:
> * Can you add a README.MD at the top level describing the purpose of the repository, explain the general technologies used (e.g. pytest), and some basic instructions how to use it?
> * Can you add some kind of requirements.txt file or something similar that helps communicate what python dependencies are required to run these tests?
> * The test output is very noisy.  I was playing with this earlier today and failed all five tests (as expected) but had a hard time seeing the test names that failed, for example.  The log output is likely useful for debugging but a more succinct test report would be nicer by default.
> * Can you add a separate readme file to the data/tpch_parquet directory explaining how these files were generated.  Also, these files end with _1 but I don't think they are scale factor 1 (are these the 0.1 files we generated?)

Yeah, I can add those things and fix some of the logging.  I'll modify the parquet file names to be 0.1 as well.

tests/common.py
gforsyth: I would recommend setting up `black` and `isort` and we can skip over any remaining questions around formatting and linting.

richtia: Thanks for the suggestion!  Just set those 2 up on my IDE

jacques-n:
Can you find a way to generate the artifacts as part of test infrastructure? Storing 20mb+ of test files is non-ideal.

gforsyth:
@richtia, in re: @jacques-n comment above, I've used this locally to generate TPC-H parquet files:
```
In [1]: import duckdb

In [2]: con = duckdb.connect()

In [3]: con.execute("CALL dbgen(sf=0.1)")
Out[3]: <duckdb.DuckDBPyConnection at 0x7fe3791a3d30>

In [4]: con.execute("EXPORT DATABASE 'parquet_out' (FORMAT PARQUET);")
Out[4]: <duckdb.DuckDBPyConnection at 0x7fe3791a3d30>

In [5]: ls parquet_out/
customer.parquet  load.sql        orders.parquet  partsupp.parquet  schema.sql
lineitem.parquet  nation.parquet  part.parquet    region.parquet    supplier.parquet
```

richtia:
> @richtia, in re: @jacques-n comment above, I've used this locally to generate TPC-H parquet files:
> 
> ```
> In [1]: import duckdb
> 
> In [2]: con = duckdb.connect()
> 
> In [3]: con.execute("CALL dbgen(sf=0.1)")
> Out[3]: <duckdb.DuckDBPyConnection at 0x7fe3791a3d30>
> 
> In [4]: con.execute("EXPORT DATABASE 'parquet_out' (FORMAT PARQUET);")
> Out[4]: <duckdb.DuckDBPyConnection at 0x7fe3791a3d30>
> 
> In [5]: ls parquet_out/
> customer.parquet  load.sql        orders.parquet  partsupp.parquet  schema.sql
> lineitem.parquet  nation.parquet  part.parquet    region.parquet    supplier.parquet
> ```

Thanks @gforsyth, I'll try this out!

richtia:
> Can you find a way to generate the test artifacts as part of test infrastructure? Storing 20mb+ of generated test files is non-ideal.

Updated.  Test data is now generated on the fly using a fixture at the start of each test case.  fixture is named prepare_tpch_parquet_data and is located in conftest.py

cc: @gforsyth

tests/verification.py
richtia: Thanks for the recommendation.  I hadn't used this operator module before.  Let me know if the change is what you were thinking.

tests/common.py
vibhatha: probably we can add a style checker somwhere down the road to check them for each PR.

tests/integration/queries/test_queries.py
```diff
@@ -0,0 +1,170 @@
...
+                    }
+                  },
+                  "names": ["L_TAX", "L_DISCOUNT"]
+                }
+              }],
+              "expectedTypeUrls": []
+            }
+            """,
+    }
+]
```
vibhatha, line 170: probably add a new line here

tests/integration/queries/tpch_sql/q1.sql
```diff
@@ -0,0 +1,21 @@
...
+from
+  '{}'
+where
+  l_shipdate <= date '1998-12-01' - interval '120' day (3)
+group by
+  l_returnflag,
+  l_linestatus
+order by
+  l_returnflag,
+  l_linestatus
```
vibhatha, line 21: same here

tests/integration/queries/tpch_sql/q2.sql
```diff
@@ -0,0 +1,36 @@
...
+            AND s_suppkey = ps_suppkey
+            AND s_nationkey = n_nationkey
+            AND n_regionkey = r_regionkey
+            AND r_name = 'EUROPE')
+ORDER BY
+    s_acctbal DESC,
+    n_name,
+    s_name,
+    p_partkey
+LIMIT 100;
```
vibhatha, line 36: same here

tests/integration/queries/tpch_sql/q3.sql
```diff
@@ -0,0 +1,21 @@
...
+    AND o_orderdate < CAST('1995-03-15' AS date)
+    AND l_shipdate > CAST('1995-03-15' AS date)
+GROUP BY
+    l_orderkey,
+    o_orderdate,
+    o_shippriority
+ORDER BY
+    revenue DESC,
+    o_orderdate
+LIMIT 10;
```
vibhatha, line 21: same here

tests/integration/queries/tpch_sql/q4.sql
```diff
@@ -0,0 +1,20 @@
...
+            *
+        FROM
+            '{}'
+        WHERE
+            l_orderkey = o_orderkey
+            AND l_commitdate < l_receiptdate)
+GROUP BY
+    o_orderpriority
+ORDER BY
+    o_orderpriority;
```
vibhatha, line 20: same here

tests/integration/queries/tpch_sql/q5.sql
```diff
@@ -0,0 +1,19 @@
...
+    AND c_nationkey = s_nationkey
+    AND s_nationkey = n_nationkey
+    AND n_regionkey = r_regionkey
+    AND r_name = 'ASIA'
+    AND o_orderdate >= CAST('1994-01-01' AS date)
+    AND o_orderdate < CAST('1995-01-01' AS date)
+GROUP BY
+    n_name
+ORDER BY
+    revenue DESC
```
vibhatha, line 19: add new a new line as mentioned before

tests/integration/queries/test_queries.py
richtia: Thanks for the catch, I added the newline to a bunch of the files.

tests/integration/queries/tpch_test_cases/all_tpch_queries.py
```diff
@@ -0,0 +1,9 @@
+from .query_1 import TESTCASE as q1
+from .query_2 import TESTCASE as q2
+from .query_3 import TESTCASE as q3
+from .query_4 import TESTCASE as q4
+from .query_5 import TESTCASE as q5
+
+TPCH_QUERY_TESTS = (
+        q1 + q2 + q3 + q4 + q5
+)
```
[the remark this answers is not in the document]
richtia: I think originally, it was because the queries include the entire substrait plan as well as the sql...but now that that's been separated out, it probably would make sense to consolidate the test cases into a single file.

tests/basetest.py
```diff
@@ -0,0 +1,54 @@
...
+
+        Parameters:
+            actual:
+                Challenge value.
+            expected:
+                Expected value to be verified against.
+            message:
+                Error message logged if objects aren't equal.
+        """
+        cls.verifier.verify_equals(actual, expected, message)
```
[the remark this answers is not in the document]
richtia: definitely not right now.  I'll remove this.

tests/integration/queries/tpch_test_cases/all_tpch_queries.py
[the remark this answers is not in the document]
gforsyth: ahh, yeah, that makes sense -- you would've been scrolling for days to  get from one test to another

tests/common.py
```diff
@@ -0,0 +1,159 @@
...
+
+class SubstraitUtils:
+    """
+    Common utility for substrait integration tests.
+    """
+
+    def __init__(self):
+        """Initializes SubstraitUtils class with a logger"""
+        self.logger = LOGGER
```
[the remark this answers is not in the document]
richtia: Just removed the logger stuff...I'll figure out all the file logging in a separate PR if needed.  The default pytest logging should be fine for now for any debugging I think.

I actually just moved one of the other functions defined outside of this class into it and a few other of the functions rely on it, so I'll keep this as a class.  I also didn't want to have to import each individual function.

tests/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+        except Exception as e:
+            print("ERROR", e)
+            pass
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_terminal_summary():
+    yield
+    print(f"Failures logged to: {FAILURES_FILE}")
+    print(f"to see run\ncat {FAILURES_FILE}")
```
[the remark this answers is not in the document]
richtia: thanks for the tip!

tests/verification.py
```diff
@@ -0,0 +1,39 @@
...
+        msg = [f"TEST FAILURE: Verifying equals: {actual} == {expected}."]
+        msg = [message] if message else msg
+
+        assert isinstance(actual, type(expected)), (
+            f"TEST FAILURE: Object types are not the same. \nActual "
+            f"type: {type(actual)}\nExpected type: {type(expected)}"
+        )
+        assert actual == expected, msg
+
+        return True
```
gforsyth, line 39: This is all looking much tidier!  Does this need to be a class?  
If it's a function, it'll save another class instantiation in your consumer test.

tests/verification.py
```diff
@@ -0,0 +1,39 @@
...
+                Object to evaluate against the expected object.
+            expected:
+                Object to be evaluated against.
+            message:
+                Message to be displayed if objects are not equal.
+
+        Returns:
+            bool:
+                True if successful, otherwise raise an error.
+        """
```
gforsyth, line 29: ahh, one more small nit -- since you're using assertions in this function, probably not much value to returning anything -- this can probably just be removed

tests/verification.py
richtia: good point...originally i had some other functions here...but now that it's just one, I'll just make it a function.  If later on it gets to a point where we have a lot more, we can make it into a class.

gforsyth (approved):
thanks for all the work on this @richtia !

README.md
```diff
@@ -0,0 +1,111 @@
...
+# Overview
+This testing repository provides instructions on how to add and run substrait consumer 
+integration tests.  These tests take in parquet files, a SQL query, and a json-formatted substrait 
+query plan as input.  The tests will run the SQL query on DuckDB, and the substrait plan on the 
+specified consumer, and then compare the two results. The test data is created using DuckDB at the 
+start of the test class using the `prepare_tpch_parquet_data` fixture, which is located in 
+`tests/conftest.py`.
+
+# Setup
+
```
westonpace, line 24:
```suggestion
```
Minor nit for consistency (alternatively, you could add a blank line after the other # headers)

README.md
```diff
@@ -0,0 +1,111 @@
...
+Create and activate your conda environment with python3.9:
```
westonpace, line 25: Does it have to be 3.9?  Or is it just that is all we are testing with internally?  I don't have any problem with 3.9, I'm just curious.

README.md
```diff
@@ -0,0 +1,111 @@
...
+```commandline
+conda create -y -n substrait_consumer_testing -c conda-forge python=3.9
+conda activate substrait_consumer_testing
+```
+
+Install requirements from `tests` directory:
+```commandline
+pip install -r requirements.txt
+```
```
westonpace, line 34:
```suggestion
Install requirements:
```commandline
pip install -r tests/requirements.txt
```
```

Or maybe you could phrase it "Install requirements (must run from the `tests` directory)

File: README.md [hunk 42]
```diff
@@ -0,0 +1,111 @@
...
+# How to Run Tests
+Tests are located in the `tests/integration` folder and are run with pytest.
+```commandline
+pytest test_acero_tpch.py
+```
+
+# Test Files
+Testing methods are located in the `tests/integration` folder and prefixed with `test_`. 
+`test_acero_tpch.py`
```
westonpace, line 43:
```suggestion
Testing methods are located in the `tests/integration` folder and prefixed with `test_`.
For example, `test_acero_tpch.py`
```

README.md
```diff
@@ -0,0 +1,111 @@
...
+Table names are determined based on the `"file_names"` in the test case args file. The order of 
+parquet file appearance in the `"file_names"` list should be consistent with the ordering for the 
+table names in the SQL query. The actual format after replacement will depend on the consumer being 
+used.
+ 
+
+# How to Add New Consumers
+Consumers should be added to the `tests/consumers` directory and provide 
+methods on how to run the substrait query plan against that consumer.  Look at 
+`tests/consumers/acero_consumer.py` for implementation details.
```
westonpace, line 111:
```suggestion
`tests/consumers/acero_consumer.py` for an example implementation.
```

tests/common.py
```diff
@@ -0,0 +1,133 @@
...
+        parquet_file_paths = self.get_full_path(file_names)
+
+        # Replace the filename placeholder in the substrait query plan with
+        # the proper parquet data file paths.
+        for count, file_path in enumerate(parquet_file_paths):
+            substrait_query = substrait_query.replace(
+                f"FILENAME_PLACEHOLDER_{count}", file_path
+            )
+
+        return pa.lib.tobytes(substrait_query)
```
westonpace, line 133:
```suggestion
        return substrait_query.encode()
```
We should probably try and avoid pyarrow dependencies outside of the consumer.  Also, is there a particular reason this is needed?  Any JSON should be representable as `UTF-8` so I think it's safe, mostly just curious.  If it's Acero that is requiring this then you might do the encoding in `AceroConsumer` just to keep things simpler.

tests/common.py
```diff
@@ -0,0 +1,133 @@
...
+        Returns:
+            List of full paths.
+        """
+        data_dir = CUR_DIR / "data" / "tpch_parquet"
+        full_paths_list = [f"{data_dir}/{dataset}" for dataset in file_names]
+
+        return full_paths_list
+
+    @staticmethod
+    def arrow_sort_tb_values(table: pa.Table, sortby: Iterable[str]) -> pa.Table:
```
westonpace, line 66: If you move this to `test_acero_tpch.py` then you can get rid of the `import pyarrow` on this file, leaving it more pure-python and consumer-agnostic.

tests/consumers/acero_consumer.py
```diff
@@ -0,0 +1,45 @@
...
+
+from pathlib import Path
+
+import pyarrow as pa
+import pyarrow.substrait as substrait
+
+
+class AceroConsumer:
+    """
+    Implementation of the acero substrait consumer Class for testing
```
westonpace, line 11:
```suggestion
    Implementation of the acero substrait consumer for testing
```

Or maybe, "adapts the Acero Substrait consumer to the test framework"?

tests/consumers/acero_consumer.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+
+    @staticmethod
+    def run_substrait_query(substrait_query: str) -> pa.Table:
```
westonpace, line 15:
```suggestion
    def run_substrait_query(substrait_query: bytes) -> pa.Table:
```
I don't know if you need `bytes` but I'm pretty sure that's what you are receiving here.

tests/consumers/acero_consumer.py
```diff
@@ -0,0 +1,45 @@
...
+            A pyarrow table resulting from running the substrait query plan.
+        """
+        buf = pa._substrait._parse_json_plan(substrait_query)
+        reader = substrait.run_query(buf)
+        result = reader.read_all()
+
+        return result
+
+    @staticmethod
+    def write_to_arrow_binary_file(table: pa.Table, filepath: Path | str) -> None:
```
westonpace, line 33: Is this used?

tests/integration/test_acero_tpch.py
```diff
@@ -0,0 +1,105 @@
...
+            sql_query:
+                SQL query.
+            substrait_query:
+                Substrait query.
+            sort_results:
+                Whether to sort the results before comparison.
+        """
+
+        # self.logger.info(f"Start to run test: {test_name}")
+        # sv.check_plan_valid(substrait_query)
```
westonpace, line 63: Remove or uncomment.

tests/integration/test_acero_tpch.py
```diff
@@ -0,0 +1,105 @@
...
+        # Reformat the sql query to be used by duck db by inserting all the
+        # parquet filepaths where the table names should be.
+        # Calculate results to verify against by running the SQL query on DuckDB
+        sql_query = self.utils.format_sql_query(sql_query, file_names)
+        duckdb_query_result_tb = self.db_connection.query(f"{sql_query}").arrow()
+
+        col_names = [x.lower() for x in subtrait_query_result_tb.column_names]
+        exp_col_names = [x.lower() for x in duckdb_query_result_tb.column_names]
+
+        # Sort results by specified column names
```
westonpace, line 79: Maybe add a comment along the lines of "we generally assume we can sort by the first column, this works for TPC-H"?

westonpace (approved):
I've got a few minor wording nits and suggestions to ease maintenance but overall I'm content with this strategy as a starting point.  I think this will evolve over time as we add additional testing and consumers so probably not best to design too much in advance.

tests/common.py
richtia: Yeah, Acero is requiring it to be bytes.  I'll do the encoding in AceroConsumer.

tests/consumers/acero_consumer.py
richtia: no, i'll remove it.

README.md
richtia: I actually moved the requirements.txt to the top level directory.  I'll update the README.

tests/consumers/acero_consumer.py
richtia: Taking your suggestion above and doing the encoding within this method.  So now string is what i'm expecting.

README.md
richtia: No, that's just what I was using. Should I reword this to something else/

tests/integration/test_acero_tpch.py
richtia: I can't say for sure sorting by the first column will work as of right now since we're still unable to run the TPC-H queries.  I was hoping to fix this up in another PR once we them running.

richtia:
@jacques-n any other modifications you'd like to see?  I think I've addressed your suggested changes and the plans here are able to run through the validator (with a few warnings).

Once this is merged, I'll submit another PR for the remaining TPCH queries, another consumer, as well as work on adding functional tests for the individual extension functions.

jacques-n:
@richtia , thanks for the ping. Thanks for working through all the feedback. I'm good with this if others who have reviewed are happy. I wish we would have seen more input from other teams working with Substrait (I prompted several) but I don't think we should hold progress for people who may or may not come along.

Thanks again. This is looking great!

richtia:
> @richtia , thanks for the ping. Thanks for working through all the feedback. I'm good with this if others who have reviewed are happy. I wish we would have seen more input from other teams working with Substrait (I prompted several) but I don't think we should hold progress for people who may or may not come along.
> 
> Thanks again. This is looking great!

Thanks @jacques-n!  I haven't made any changes since @westonpace and @gforsyth gave their approvals, so I think we should be okay to merge this.  If anyone else has issues with the tests here, i'm happy to address their feedback in additional patches as well.

vibhatha (approved):

PR #3, opened 2022-08-24
feat: add duckdb consumer and tpch tests

richtia:
PR to add tpch query tests using DuckDB as a consumer

### DIFF
--- a/tests/consumers/duckdb_consumer.py
+++ b/tests/consumers/duckdb_consumer.py
@@ -1,0 +1,64 @@
+import string
+from pathlib import Path
+from typing import Iterable
+
+import pyarrow as pa
+
+from ..common import SubstraitUtils
+
+
+class DuckDBConsumer:
+    """
+    Implementation of the DuckDB substrait consumer Class for testing
+    """
+
+    def __init__(self, db_connection):
+        """
+        Initialize the DuckDBConsumer Class with a duckdb connection.
+        """
+        self.db_connection = db_connection
+
+    def run_substrait_query(self, sql_query: str) -> pa.Table:
+        """
+        Convert the SQL query into a substrait query plan and run the plan against DuckDB.
+
+        Parameters:
+            sql_query:
+                A SQL query string.
+
+        Returns:
+            A pyarrow table resulting from running the substrait query plan.
+        """
+        duckdb_substrait_plan = self.db_connection.get_substrait(sql_query)
+        proto_bytes = duckdb_substrait_plan.fetchone()[0]
+        return self.db_connection.from_substrait(proto_bytes).arrow()
+
+    def load_tables_from_parquet(
+        self,
+        created_tables: set,
+        file_names: Iterable[str],
+    ) -> list:
+        """
+        Load all the parquet files into separate tables in DuckDB.
+
+        Parameters:
+            created_tables:
+                The set of tables that have already been created.
+            file_names:
+                Name of parquet files.
+
+        Returns:
+            A list of the table names.
+        """
+        parquet_file_paths = SubstraitUtils.get_full_path(file_names)
+        table_names = []
+        for file_name, file_path in zip(file_names, parquet_file_paths):
+            table_name = Path(file_name).stem
+            table_name = table_name.translate(str.maketrans("", "", string.punctuation))
+            if table_name not in created_tables:
+                create_table_sql = f"CREATE TABLE {table_name} AS SELECT * FROM read_parquet('{file_path}');"
+                self.db_connection.execute(create_table_sql)
+                created_tables.add(table_name)
+            table_names.append(table_name)
+
+        return table_names
--- a/tests/integration/test_duckdb_tpch.py
+++ b/tests/integration/test_duckdb_tpch.py
@@ -1,0 +1,89 @@
+import duckdb
+import pytest
+
+from ..consumers import DuckDBConsumer
+from ..parametrization import custom_parametrization
+from ..verification import verify_equals
+from .queries.tpch_test_cases import TPCH_QUERY_TESTS
+
+
+@pytest.mark.usefixtures("prepare_tpch_parquet_data")
+class TestDuckDBConsumer:
+    """
+    Test Class for testing Substrait using DuckDB as a consumer.
+    """
+
+    @staticmethod
+    @pytest.fixture(scope="class", autouse=True)
+    def setup_teardown_class(request):
+        cls = request.cls
+
+        cls.db_connection = duckdb.connect()
+        cls.db_connection.execute("INSTALL substrait")
+        cls.db_connection.execute("LOAD substrait")
+        cls.consumer = DuckDBConsumer(cls.db_connection)
+        cls.created_tables = set()
+
+        yield
+
+        cls.db_connection.close()
+
+    @custom_parametrization(TPCH_QUERY_TESTS)
+    def test_substrait_query(
+        self,
+        test_name: str,
+        file_names: list,
+        sql_query: str,
+        substrait_query: str,
+        sort_results: bool = False,
+    ) -> None:
+        """
+        1.  Load all the parquet files into DuckDB as separate tables.
+        2.  Format the SQL query to work with DuckDB by inserting all the table names.
+        3.  Execute the SQL on DuckDB.
+        4.  Run the substrait query plan.
+        5.  Compare substrait query plan results against the results of
+            running the SQL on DuckDB.
+
+        Parameters:
+            test_name:
+                Name of test.
+            file_names:
+                List of parquet files.
+            sql_query:
+                SQL query.
+        """
+
+        # Load the parquet files into DuckDB and return all the table names as a list
+        table_names = self.consumer.load_tables_from_parquet(
+            self.created_tables, file_names
+        )
+
+        # Format the sql query by inserting all the table names
+        sql_query = sql_query.format(*table_names)
+
+        # Convert the SQL into a substrait query plan and run the plan.
+        subtrait_query_result_tb = self.consumer.run_substrait_query(sql_query)
+
+        # Calculate results to verify against by running the SQL query on DuckDB
+        duckdb_sql_result_tb = self.db_connection.query(f"{sql_query}").arrow()
+
+        col_names = [x.lower() for x in subtrait_query_result_tb.column_names]
+        exp_col_names = [x.lower() for x in duckdb_sql_result_tb.column_names]
+
+        # Verify results between substrait plan query and sql running against
+        # duckdb are equal.
+        verify_equals(
+            col_names,
+            exp_col_names,
+            message=f"Actual column names: \n{col_names} \n"
+            f"are not equal to the expected"
+            f"column names: \n{exp_col_names}",
+        )
+        verify_equals(
+            subtrait_query_result_tb,
+            duckdb_sql_result_tb,
+            message=f"Result table: \n{subtrait_query_result_tb} \n"
+            f"is not equal to the expected "
+            f"table: \n{duckdb_sql_result_tb}",
+        )
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# PyCharm
+.idea/
--- a/tests/consumers/__init__.py
+++ b/tests/consumers/__init__.py
@@ -1,0 +1,2 @@
+from tests.consumers.acero_consumer import AceroConsumer
+from tests.consumers.duckdb_consumer import DuckDBConsumer
--- a/tests/consumers/acero_consumer.py
+++ b/tests/consumers/acero_consumer.py
@@ -10,18 +10,18 @@ class AceroConsumer:
     """
 
     @staticmethod
-    def run_substrait_query(substrait_query: str) -> pa.Table:
+    def run_substrait_query(substrait_query: bytes) -> pa.Table:
         """
         Run the given substrait query and return the result
 
         Parameters:
-            substrait_query: A json formatted string representing the
+            substrait_query: A json formatted byte representation of the
             substrait query plan
 
         Returns:
             A pyarrow table resulting from running the substrait query plan.
         """
-        buf = pa._substrait._parse_json_plan(substrait_query.encode())
+        buf = pa._substrait._parse_json_plan(substrait_query)
         reader = substrait.run_query(buf)
         result = reader.read_all()
 
--- a/tests/integration/queries/tpch_sql/q1.sql
+++ b/tests/integration/queries/tpch_sql/q1.sql
@@ -12,7 +12,7 @@ SELECT
 FROM
   '{}'
 WHERE
-  l_shipdate <= date '1998-12-01' - interval '120' day (3)
+  l_shipdate <= date '1998-12-01' - interval '120' day
 GROUP BY
   l_returnflag,
   l_linestatus
--- a/tests/integration/test_acero_tpch.py
+++ b/tests/integration/test_acero_tpch.py
@@ -6,7 +6,7 @@ import pytest
 from pyarrow import compute
 
 from ..common import SubstraitUtils
-from ..consumers.acero_consumer import AceroConsumer
+from ..consumers import AceroConsumer, DuckDBConsumer
 from ..parametrization import custom_parametrization
 from ..verification import verify_equals
 from .queries.tpch_test_cases import TPCH_QUERY_TESTS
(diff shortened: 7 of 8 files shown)


tests/integration/test_duckdb_tpch.py
```diff
@@ -0,0 +1,108 @@
...
+            self.db_connection, self.created_tables, file_names, parquet_file_paths
+        )
+
+        # Format the sql query by inserting all the table names
+        sql_query = sql_query.format(*table_names)
+
+        # Convert the SQL into a substrait query plan and run the plan.
+        subtrait_query_result_tb = self.consumer.run_substrait_query(
+            self.db_connection, sql_query
+        )
```
pdet, line 74: It seems this function is DuckDB consuming its own produced substrait plans, right?

If that's the case (and goal), you can simplify the code by setting `PRAGMA enable_verification`.

Basically, when that pragma is set, DuckDB will internally execute the query and the substrait plan resulting from `get_substrait(SQL)` and compare it's resulting throwing an Error if they do not match.

I think the verify_equals is more interesting if you are comparing results of two different systems consuming a substrait plan.

tests/integration/test_duckdb_tpch.py
```diff
@@ -0,0 +1,108 @@
...
+        # Sort results by specified column names
+        if sort_results:
+            subtrait_sort_col = subtrait_query_result_tb.column_names[0]
+            subtrait_query_result_tb = self.utils.arrow_sort_tb_values(
+                subtrait_query_result_tb, sortby=[subtrait_sort_col]
+            )
+            duckdb_sort_col = duckdb_sql_result_tb.column_names[0]
+            duckdb_sql_result_tb = self.utils.arrow_sort_tb_values(
+                duckdb_sql_result_tb, sortby=[duckdb_sort_col]
+            )
```
pdet, line 91: Just to clarify, I do this sorting on column names, because internally we do test the consumption of ibis substrait plans vs ibis query results, and ibis does not output the results in the order set by the SQL query. Hence, this does not seem to be necessary here.

tests/consumers/duckdb_consumer.py
```diff
@@ -0,0 +1,64 @@
...
+    @staticmethod
+    def run_substrait_query(db_connection, sql_query: str) -> pa.Table:
+        """
+        Convert the SQL query into a substrait query plan and run the plan against DuckDB.
+
+        Parameters:
+            db_connection:
+                DuckDB connection.
+            sql_query:
+                A json formatted string representing the substrait query plan.
```
pdet, line 22: What does json formatted string representing the substrait query plan means in this context? Isn't this just a plain SQL query string?

tests/integration/test_duckdb_tpch.py
richtia: > It seems this function is DuckDB consuming its own produced substrait plans, right?
> 
> If that's the case (and goal), you can simplify the code by setting `PRAGMA enable_verification`.
> 
> Basically, when that pragma is set, DuckDB will internally execute the query and the substrait plan resulting from `get_substrait(SQL)` and compare it's resulting throwing an Error if they do not match.
> 
> I think the verify_equals is more interesting if you are comparing results of two different systems consuming a substrait plan.

Yes, that's correct.  However, the eventual plan is to have tests where DuckDB is consuming substrait plans produced by other producers as well.

tests/consumers/duckdb_consumer.py
richtia: Thanks for catching this, i'll update it.

pdet:
LGTM

tests/consumers/duckdb_consumer.py
```diff
@@ -0,0 +1,64 @@
...
+                Name of parquet files.
+            parquet_file_paths:
+                Full path of parquet files.
+
+        Returns:
+            A list of the table names.
+        """
+        table_names = []
+        for file_name, file_path in zip(file_names, parquet_file_paths):
+            table_name = os.path.splitext(file_name)[0]
```
gforsyth, line 56: small nit -- I think we settled on `pathlib` in the initial commit -- do we want to keep that consistent?  I believe you can use the `stem` or `suffix` attributes of a `pathlib.Path` in place of `splitext()`

tests/integration/test_duckdb_tpch.py
```diff
@@ -0,0 +1,95 @@
...
+from ..consumers.duckdb_consumer import DuckDBConsumer
+from ..parametrization import custom_parametrization
+from ..verification import verify_equals
+from .queries.tpch_test_cases import TPCH_QUERY_TESTS
+
+
+@pytest.mark.usefixtures("prepare_tpch_parquet_data")
+class TestDuckDBConsumer:
+    """
+    Test Class for testing Substrait using Acero as a consumer.
```
gforsyth, line 14:
```suggestion
    Test Class for testing Substrait using DuckDB as a consumer.
```

tests/integration/test_duckdb_tpch.py
```diff
@@ -0,0 +1,95 @@
...
+        Parameters:
+            test_name:
+                Name of test.
+            file_names:
+                List of parquet files.
+            sql_query:
+                SQL query.
+        """
+
+        parquet_file_paths = self.utils.get_full_path(file_names)
```
gforsyth, line 59: Since you have this utility, maybe it should be rolled into the `consumer` class?  That way you only have to pass in the `file_names` (and then have the `consumer` class figure out the paths using your helper method)

tests/consumers/duckdb_consumer.py
```diff
@@ -0,0 +1,64 @@
...
+        Returns:
+            A pyarrow table resulting from running the substrait query plan.
+        """
+        duckdb_substrait_plan = db_connection.get_substrait(sql_query)
+        proto_bytes = duckdb_substrait_plan.fetchone()[0]
+        return db_connection.from_substrait(proto_bytes).arrow()
+
+    @staticmethod
+    def load_tables_from_parquet(
+        db_connection,
```
gforsyth, line 33: Both of these methods need the `db_connection` -- it could be kept with the `DuckDBConsumer` class and shoved into an attribute so it isn't required to be passed to both static methods (which would no longer be static)

gforsyth (changes requested):
A few small nits and a few questions, nothing major.  Otherwise looks good!

tests/consumers/duckdb_consumer.py
richtia: Yeah, we did.  Just updated!

tests/integration/test_duckdb_tpch.py
richtia: Good idea.  Done.

tests/consumers/duckdb_consumer.py
richtia: Thanks for the suggestion, this is cleaner!

gforsyth (approved):
Nice!  Looks good to me.

tests/integration/test_duckdb_tpch.py
```diff
@@ -0,0 +1,89 @@
...
+    def setup_teardown_class(request):
+        cls = request.cls
+
+        cls.db_connection = duckdb.connect()
+        cls.db_connection.execute("INSTALL substrait")
+        cls.db_connection.execute("LOAD substrait")
+        cls.consumer = DuckDBConsumer(cls.db_connection)
+        cls.created_tables = set()
+
+        yield
```
westonpace, line 27: What is this yield here for?

westonpace (commented):
I'm a little confused what the goal is here.  For Acero, our test corpus consists of `SQL` and `Substrait Query` and we have:

Test: Substrait Query -> Acero -> Arrow
Oracle: SQL -> DuckDb -> Arrow

However, in this PR, we have:

Test: SQL -> DuckDb -> Substrait Query -> DuckDb -> Arrow
Oracle: SQL -> DuckDb -> Arrow

In other words, DuckDb is not running against the same Substrait Query that other consumers (e.g. Acero) are running against.  The `substrait_query` parameter to `test_substrait_query` is completely ignored.

richtia:
>

richtia:
> I'm a little confused what the goal is here. For Acero, our test corpus consists of `SQL` and `Substrait Query` and we have:
> 
> Test: Substrait Query -> Acero -> Arrow Oracle: SQL -> DuckDb -> Arrow
> 
> However, in this PR, we have:
> 
> Test: SQL -> DuckDb -> Substrait Query -> DuckDb -> Arrow Oracle: SQL -> DuckDb -> Arrow
> 
> In other words, DuckDb is not running against the same Substrait Query that other consumers (e.g. Acero) are running against. The `substrait_query` parameter to `test_substrait_query` is completely ignored.

Ahh...I see your confusion.  I should have been more clear.

The reason for the disconnect right now is because the manually generated substrait queries in the Acero tests are being executed against local parquet files, which DuckDB doesn't support yet (I've opened an issue for them to add support here: https://github.com/duckdblabs/substrait/issues/3).  
Whereas DuckDB currently only supports producing/running substrait plans that use named tables (which Acero doesn't support yet.  Vibhatha has the PR in review: https://github.com/apache/arrow/pull/14024/).  

Once there is the same provider support in Acero/DuckDB, I'll open another PR to add the other tests for each consumer so that there is parity in the testing.

The plan is to have: 
Acero Consumer Tests:
1. Isthmus Provider: Substrait Query (local parquet)-> `Acero` -> Arrow (currently supported)
2. DuckDb Provider: SQL -> DuckDb -> Substrait Query (named tables) -> `Acero` -> Arrow (currently not supported)

DuckDB Consumer Tests:
1. Isthmus Provider (using local parquet): Substrait Query (local parquet) -> `DuckDB` -> Arrow (currently not supported)
2. DuckDb Provider (using named tables): SQL -> DuckDb -> Substrait Query (named tables) -> `DuckDB` -> Arrow (currently supported)

I was going to updated the README once the support is all there...but I can also just update with the existing support in this PR if that helps avoid any confusion too.

richtia:
Accidentally closed the PR...reopening

tests/integration/test_duckdb_tpch.py
richtia: the yield is for the `setup_teardown_class` fixture.

Once all the tests are run, anything after this yield is called.  In this specific case, we close the duckdb connection at the end.

westonpace:
> Once there is the same provider support in Acero/DuckDB, I'll open another PR to add the other tests for each consumer so that there is parity in the testing.

This seems like a reasonable approach.

jacques-n:
It's seems based on comments that this patch doesn't integrates with neither acero nor isthmus. I think it should integrate with at least one before being merged. It sounds like thst would be isthmus given the state of things?

westonpace:
> It's seems based on comments that this patch doesn't integrates with neither acero nor isthmus. I think it should integrate with at least one before being merged. It sounds like thst would be isthmus given the state of things?

I'm not entirely sure that is needed.  It seems like we have four tasks planned.

| Id | Producer | Consumer |
| --- | --- | --- |
| A | Isthmus | Acero |
| B | DuckDb | Acero |
| C | Isthmus | DuckDb |
| D | DuckDb | DuckDb |

Task A has already merged.
Task B is blocked by Acero work.
Task C is blocked by DuckDb work.
Task D is added by this PR.

jacques-n:
Why would D be done here and not in duckdb? It sounds like an internal test of duckdb, not an integration test of two different systems.

richtia:
> Why would D be done here and not in duckdb? It sounds like an internal test of duckdb, not an integration test of two different systems.

D was done mostly as an example to demonstrate how we should use duckdb as a producer or consumer in an integration test with other producers/consumers.  It's modular enough so that we can swap out consumers/producers relatively easily. I think it would be good to keep these here until we have other consumer/producer integration tests with the duckdb producer/consumer.

jacques-n:
> D was done mostly as an example to demonstrate how we should use duckdb as a producer or consumer in an integration test with other producers/consumers. It's modular enough so that we can swap out consumers/producers relatively easily. I think it would be good to keep these here until we have other consumer/producer integration tests with the duckdb producer/consumer.

Examples are super useful. However, I'm not convinced this is the place for examples. How about we put it on an example page on the Substrait website or in a Substrait examples repo. 

Integration testing is sacred from my pov. It's what proves that this isn't a single-system standard. Having a single system talking to itself that may be entirely inconsistent from other systems talking to each other in the same repo feels like it sends the wrong message. It suggests that we're doing integration testing with DuckDB when we aren't (yet).

Given both your pushback, it seems like this is more controversial than I would expect it to be. Maybe we had different intentions for this repo or I'm missing part of the picture that you both have in your heads?

westonpace:
I think the idea is that there are P different producers and Q different queries.  For each consumer you will have `P * Q` different test cases:

```
q_1 x p_1
q_1 x p_2
...
q_1 x p_P
q_2 x p_1
...
q_Q x p_P
```

Eventually that won't be exact because some queries will only make sense for a subset of producers.  Also, today all queries are SQL but I think there may be a future where that isn't the case.

However, the point is that, for each query, a producer should return the correct answer, regardless of the producer that is converting query to plan.  You could imagine a support table:

| Consumer | Isthmus | DuckDb | Ibis | ... | dplyr |
| --- | --- | --- | --- | --- | --- |
| Acero | :heavy_check_mark: | :x: | :heavy_check_mark:  | ... | :heavy_check_mark: |
| DuckDb | :heavy_check_mark: | :heavy_check_mark: | :x: | ... | :heavy_check_mark: |

DuckDb is a producer, so there are a set of test cases created from it:

```
q_1 x DuckDb
q_2 x DuckDb
...
q_Q x DuckDb
```

We could call these tests "Integration" and skip them.  Yielding:

| Consumer | Isthmus | DuckDb | Ibis | ... | dplyr |
| --- | --- | --- | --- | --- | --- |
| Acero | :heavy_check_mark: | :x: | :heavy_check_mark:  | ... | :heavy_check_mark: |
| DuckDb | :heavy_check_mark: | (NA) | :x: | ... | :heavy_check_mark: |

...however it's not clear to me that we have to skip these cases.  If it isn't too much trouble to include them then it seems it will keep things simpler rather than confuse users why there might be an NA in the table.

westonpace:
"Support table" is maybe the wrong word.  "Test result table" might be more descriptive.  In theory a row should have check marks across the board or X's across the board.  In practice we've seen that each producer's nuances tend to lead to scattered X's.  However, this is not something we should be advertising or telling users, which is why "support table" is not really the right word.

westonpace:
> It suggests that we're doing integration testing with DuckDB when we aren't (yet).

I agree that adding this PR is not completely satisfying "add duckdb consumer and tpch tests".  That title would imply to me that we are fully testing duckdb against all possible producers.

Perhaps we can reword the PR title to "add duckdb consumer and tpch tests for duckdb producer" and then make sure we have github issues for "add duckdb consumer and tpch tests for Isthmus producer" and "add acero consumer and tpch tests for duckdb producer".

jacques-n:
I don't like merging something that may have a completely disjunct perspective on what a Substrait plan is. At least early on, let's only merge things here that also have a "trust chain". Once duckdb is in the trust chain via acero and/or isthmus, we can also add the inside-duck connection.

pdet:
Hey guys, 

so for a bit of context, what tpc-h queries can Acero and Isthmus round-trip?

From what I can see in the repo, acero is a consumer-only engine, but I'm not exactly sure how to consume/produce substrait plans from isthmus (I also haven't heard about isthmus before, so any ref pointers would be nice).

If these systems are (or are close to) a ground truth of accepted substrait plans, I'm of course happy to add them to our own substrait CI (We already have a bit on Ibis), since the acero consumption code is in this repo, having it in our CI and going from there should be pretty straightforward, and I can start working on adapting to it. :-).

richtia:
> Hey guys,
> 
> so for a bit of context, what tpc-h queries can Acero and Isthmus round-trip?
> 
> From what I can see in the repo, acero is a consumer-only engine, but I'm not exactly sure how to consume/produce substrait plans from isthmus (I also haven't heard about isthmus before, so any ref pointers would be nice).
> 
> If these systems are (or are close to) a ground truth of accepted substrait plans, I'm of course happy to add them to our own substrait CI (We already have a bit on Ibis), since the acero consumption code is in this repo, having it in our CI and going from there should be pretty straightforward, and I can start working on adapting to it. :-).

Hey Pedro, here's the repo with Isthmus: https://github.com/substrait-io/substrait-java/tree/main/isthmus

The TPC-H substrait plans in this repo were generated using Isthmus and then modified to read data from parquet files since that's is what Acero currently support:
https://github.com/substrait-io/consumer-testing/tree/main/tests/integration/queries/tpch_substrait_plans
(I know there is a plan for duckdb to add that support: https://github.com/duckdblabs/substrait/issues/3).  

By default Isthmus has named tables in the substrait plans though, similar to duckdb produced substrait plans, so you may be able to run those already.

westonpace:
Another good idea might be to run the DuckDb (or whatever producer we are testing) plans through the validator.

pdet:
Hi @westonpace, we already do that in the DuckDB CI, all tpc-h queries go through the validator.

Feel free to adapt the code https://github.com/duckdblabs/substrait/blob/main/test/python/test_validator.py

richtia:
Thanks @pdet
Just added a test to check the validity of substrait plans generated via isthmus as well as duckdb: 
https://github.com/substrait-io/consumer-testing/pull/3/files#diff-41f46f832fcb78dd7d4458cc161defc9bf5e50da0e9e0f570e868566f0b43d69

Looks like Isthmus only need warnings to be overridden, but duckdb needs a few errors to be overridden to pass the validator.

PR #6, opened 2022-11-08
Add Isthmus producer

richtia:
Add isthmus consumer to integration testing framework

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,0 @@
-all:
-	cd ${SUBSTRAIT_JAVA_HOME}/isthmus; ../gradlew shadowJar
-	mkdir -p jars
-	cp ${SUBSTRAIT_JAVA_HOME}/isthmus/build/libs/*all.jar ./jars/
-
-clean:
-	rm -r jars
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,7 @@
+all:
+	cd ${SUBSTRAIT_JAVA_HOME}/isthmus; ../gradlew shadowJar
+	mkdir -p jars
+	cp ${SUBSTRAIT_JAVA_HOME}/isthmus/build/libs/*all.jar ./jars/
+
+clean:
+	rm -r jars
--- a/README.md
+++ b/README.md
@@ -32,8 +32,8 @@ conda create -y -n substrait_consumer_testing -c conda-forge python=3.9 openjdk
 conda activate substrait_consumer_testing
 ```
 *Note: Java is used by Jpype to access the Isthmus producer.  
-JPype has been tested with Java versions from Java 1.7 to Java 13.
-However, it should work with all versions of Java.*
+JPype should work with all versions of Java but to see details on which versions are 
+officially supported see https://jpype.readthedocs.io/en/latest/install.html*
 
 Install requirements from the top level directory:
 ```commandline
@@ -41,12 +41,14 @@ pip install -r requirements.txt
 ```
 
 Get the java dependencies needed by the Isthmus Substrait producer:
+1. Clone the substrait-java repo
+2. From the consumer-testing repo, run the build-and-copy-isthmus-shadow-jar.sh script
 ```commandline
 git clone https://github.com/substrait-io/substrait-java.git
-export SUBSTRAIT_JAVA_HOME=$(pwd)/substrait-java
 cd consumer-testing
-make
+sh build-and-copy-isthmus-shadow-jar.sh
 ```
+*This shell script may not work on Windows environments.*
 
 # How to Run Tests
 TPCH tests are located in the `tests/integration` folder and substrait function tests
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ which is located in `tests/conftest.py`.
 # Setup
 Create and activate your conda environment with python3.9:
 ```commandline
-conda create -y -n substrait_consumer_testing -c conda-forge python=3.9
+conda create -y -n substrait_consumer_testing -c conda-forge python=3.9 openjdk
 conda activate substrait_consumer_testing
 ```
 
--- a/README.md
+++ b/README.md
@@ -31,6 +31,9 @@ Create and activate your conda environment with python3.9:
 conda create -y -n substrait_consumer_testing -c conda-forge python=3.9 openjdk
 conda activate substrait_consumer_testing
 ```
+*Note: Java is used by Jpype to access the Isthmus producer.  
+JPype has been tested with Java versions from Java 1.7 to Java 13.
+However, it should work with all versions of Java.*
 
 Install requirements from the top level directory:
 ```commandline
--- a/README.md
+++ b/README.md
@@ -39,6 +39,15 @@ Install requirements from the top level directory:
 ```commandline
 pip install -r requirements.txt
 ```
+
+Get the java dependencies needed by the Isthmus Substrait producer:
+```commandline
+git clone https://github.com/substrait-io/substrait-java.git
+export SUBSTRAIT_JAVA_HOME=$(pwd)/substrait-java
+cd consumer-testing
+make
+```
+
 # How to Run Tests
 TPCH tests are located in the `tests/integration` folder and substrait function tests
 are located in the `tests/functional` folder.
--- a/tests/context.py
+++ b/tests/context.py
@@ -1,4 +1,11 @@
-from tests.java_definitions import *
+from pathlib import Path
+
+import jpype.imports
+
+import tests.java_definitions as java
+
+REPO_DIR = Path(__file__).parent.parent
+from com.google.protobuf.util import JsonFormat as json_formatter
 
 schema_file = Path.joinpath(REPO_DIR, "tests/data/tpch_parquet/schema.sql")
 
@@ -16,7 +23,7 @@ def produce_isthmus_substrait(sql_string, schema_list):
     Returns:
         Substrait plan in json format.
     """
-    sql_to_substrait = SqlToSubstraitClass()
+    sql_to_substrait = java.SqlToSubstraitClass()
     java_sql_string = jpype.java.lang.String(sql_string)
     plan = sql_to_substrait.execute(java_sql_string, schema_list)
     json_plan = json_formatter.printer().print_(plan)
@@ -34,17 +41,17 @@ def get_schema(file_names):
     Returns:
         List of all schemas as a java list.
     """
-    arr = ArrayListClass()
+    arr = java.ArrayListClass()
     if file_names:
         text_schema_file = open(schema_file)
         schema_string = text_schema_file.read().replace("\n", " ").split(";")[:-1]
         for create_table in schema_string:
-            java_obj = JObject @ JString(create_table)
+            java_obj = jpype.JObject @ jpype.JString(create_table)
             arr.add(java_obj)
     else:
-        java_obj = JObject @ JString(
+        java_obj = jpype.JObject @ jpype.JString(
             "CREATE TABLE T(a integer, b integer, c boolean, d boolean)"
         )
         arr.add(java_obj)
 
-    return ListClass @ arr
+    return java.ListClass @ arr
--- a/tests/context.py
+++ b/tests/context.py
@@ -1,0 +1,77 @@
+import os
+from pathlib import Path
+
+import jpype
+import jpype.imports
+from jpype.types import *
+
+REPO_DIR = Path(__file__).parent.parent
+schema_file = Path.joinpath(REPO_DIR, "tests/data/tpch_parquet/schema.sql")
+
+calcite_jars = Path.joinpath(REPO_DIR, "jars/*")
+
+the_java_home = "CONDA_PREFIX"
+if "JAVA_HOME" in os.environ:
+    the_java_home = "JAVA_HOME"
+
+java_home_path = os.environ[the_java_home]
+jvm_path = java_home_path
+
+if not os.path.isfile(jvm_path):
+    jvm_path = java_home_path + "/lib/libjli.dylib"
+
+jpype.startJVM("--enable-preview", convertStrings=True, jvmpath=jvm_path)
+jpype.addClassPath(calcite_jars)
+from com.google.protobuf.util import JsonFormat
+
+ArrayListClass = jpype.JClass("java.util.ArrayList")
+ListClass = jpype.JClass("java.util.List")
+SqlToSubstraitClass = jpype.JClass("io.substrait.isthmus.SqlToSubstrait")
+json_formatter = JsonFormat
+
+
+def produce_isthmus_substrait(sql_string, schema_list):
+    """
+    Produce the substrait plan using Isthmus.
+
+    Parameters:
+        sql_string:
+            SQL query.
+        schema_list:
+            List of schemas.
+
+    Returns:
+        Substrait plan in json format.
+    """
+    sql_to_substrait = SqlToSubstraitClass()
+    java_sql_string = jpype.java.lang.String(sql_string)
+    plan = sql_to_substrait.execute(java_sql_string, schema_list)
+    json_plan = json_formatter.printer().print_(plan)
+    return json_plan
+
+
+def get_schema(file_names):
+    """
+    Create the list of schemas based on the given file names.  If there are no files
+    give, a custom schema for the data is used.
+
+    Parameters:
+        file_names: List of file names.
+
+    Returns:
+        List of all schemas as a java list.
+    """
+    arr = ArrayListClass()
+    if file_names:
+        text_schema_file = open(schema_file)
+        schema_string = text_schema_file.read().replace("\n", " ").split(";")[:-1]
+        for create_table in schema_string:
+            java_obj = JObject @ JString(create_table)
+            arr.add(java_obj)
+    else:
+        java_obj = JObject @ JString(
+            "CREATE TABLE T(a integer, b integer, c boolean, d boolean)"
+        )
+        arr.add(java_obj)
+
+    return ListClass @ arr
--- a/tests/context.py
+++ b/tests/context.py
@@ -1,33 +1,6 @@
-import os
-from pathlib import Path
+from tests.java_definitions import *
 
-import jpype
-import jpype.imports
-from jpype.types import *
-
-REPO_DIR = Path(__file__).parent.parent
 schema_file = Path.joinpath(REPO_DIR, "tests/data/tpch_parquet/schema.sql")
-
-calcite_jars = Path.joinpath(REPO_DIR, "jars/*")
-
-the_java_home = "CONDA_PREFIX"
-if "JAVA_HOME" in os.environ:
-    the_java_home = "JAVA_HOME"
-
-java_home_path = os.environ[the_java_home]
-jvm_path = java_home_path
-
-if not os.path.isfile(jvm_path):
-    jvm_path = java_home_path + "/lib/libjli.dylib"
-
-jpype.startJVM("--enable-preview", convertStrings=True, jvmpath=jvm_path)
-jpype.addClassPath(calcite_jars)
-from com.google.protobuf.util import JsonFormat
-
-ArrayListClass = jpype.JClass("java.util.ArrayList")
-ListClass = jpype.JClass("java.util.List")
-SqlToSubstraitClass = jpype.JClass("io.substrait.isthmus.SqlToSubstrait")
-json_formatter = JsonFormat
 
 
 def produce_isthmus_substrait(sql_string, schema_list):
--- a/tests/functional/common.py
+++ b/tests/functional/common.py
@@ -81,12 +81,12 @@ def substrait_function_test(
                 sql_query, consumer, ibis_expr(*args)
             )
         else:
-            pytest.skip("ibis expression currently undefined")
+            pytest.xfail("ibis expression currently undefined")
     else:
         if type(producer) in supported_producers:
             substrait_plan = producer.produce_substrait(sql_query, consumer)
         else:
-            pytest.skip(f"{type(producer).__name__} does not support the following SQL: "
+            pytest.xfail(f"{type(producer).__name__} does not support the following SQL: "
                         f"{sql_query}")
 
     actual_result = consumer.run_substrait_query(substrait_plan)
--- a/tests/functional/common.py
+++ b/tests/functional/common.py
@@ -97,7 +97,7 @@ def substrait_function_test(
 
 
 def load_custom_duckdb_table(db_connection):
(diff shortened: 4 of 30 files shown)


tests/context.py
```diff
@@ -0,0 +1,77 @@
...
+from pathlib import Path
+
+import jpype
+import jpype.imports
+from jpype.types import *
+
+REPO_DIR = Path(__file__).parent.parent
+schema_file = Path.joinpath(REPO_DIR, "tests/data/tpch_parquet/schema.sql")
+
+calcite_jars = Path.joinpath(REPO_DIR, "jars/*")
```
aucahuasi, line 11: I would encapsulate these global definitions (lines 11-24) into a separate python file/module.  So, in the future we can change -if we want/need to- the integration layer we are using to another thing that is not jpype.

README.md
```diff
@@ -28,7 +28,7 @@ which is located in `tests/conftest.py`.
 # Setup
 Create and activate your conda environment with python3.9:
 ```commandline
-conda create -y -n substrait_consumer_testing -c conda-forge python=3.9
+conda create -y -n substrait_consumer_testing -c conda-forge python=3.9 openjdk
```
aucahuasi, line 31: I'll check if we need to pin the openjdk version (sometimes we need specific versions to support these java integrations)

aucahuasi (changes requested):
Would be great to separate the jpype definitions from context.py (as pointed out in the comment)

richtia:
> Would be great to separate the jpype definitions from context.py (as pointed out in the comment)

Done! Thanks for the suggesstion.

richtia:
I ended up grouping the SQL query param up with the list of supported producers as a tuple...but i'm open to just having 2 separate test case arguments for that too.

README.md
aucahuasi: I did a quick check for the JPype dependency and read their setup files; I could not spot anything regarding a specific Java version.  The [documentation says it was tested with Java versions 1.7 to 13](https://jpype.readthedocs.io/en/latest/install.html), so I guess we can recommend those versions to the users without pinning.   Also, the conda package for JPype is not pinning java here ([they don't even include java as part of the dependencies in the yaml definition](https://github.com/conda-forge/jpype1-feedstock/blob/main/recipe/meta.yaml))

richtia: Thanks for checking this!  I'l update the readme with this info.

tests/java_definitions.py
```diff
@@ -0,0 +1,28 @@
...
+    jvm_path = java_home_path + "/lib/libjli.dylib"
+
+jpype.startJVM("--enable-preview", convertStrings=True, jvmpath=jvm_path)
+jpype.addClassPath(calcite_jars)
+from com.google.protobuf.util import JsonFormat
+
+ArrayListClass = jpype.JClass("java.util.ArrayList")
+ListClass = jpype.JClass("java.util.List")
+SqlToSubstraitClass = jpype.JClass("io.substrait.isthmus.SqlToSubstrait")
+json_formatter = JsonFormat
```
westonpace, line 28: This seems odd.  I'm not sure we need to alias and reexport something from protobuf.  Can't the places that need this do their own `from com.google.protobuf.util import JsonFormat as json_formatter`? (although the alias is fishy too, is this a class or a method?)

tests/context.py
```diff
@@ -0,0 +1,50 @@
+from tests.java_definitions import *
```
westonpace, line 1: I'm not a big fan of the * import.  How about:

```
import tests.java_definitions as java
...
sql_to_substrait = java.SqlToSubstraitClass()
```

Or you could do `isthmus` instead of `java`:

```
import tests.java_definitions as isthmus
...
sql_to_substrait = isthmus.SqlToSubstraitClass()
```

tests/java_definitions.py
```diff
@@ -0,0 +1,28 @@
+import os
+from pathlib import Path
+
+import jpype
+import jpype.imports
+from jpype.types import *
```
westonpace, line 6: I'm ok with this * import as it is self-contained though I still think it'd be cleaner without.

westonpace (changes requested):
Great work and thanks!  A few nits but nothing major.

tests/functional/common.py
```diff
@@ -97,7 +102,7 @@ def substrait_function_test(
 
 
 def load_custom_duckdb_table(db_connection):
-    db_connection.execute("create table t (a int, b int, c boolean, d boolean)")
+    db_connection.execute("create table t (a BIGINT, b BIGINT, c boolean, d boolean)")
```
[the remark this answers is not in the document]
richtia: DuckDB int is actually int32 and isthmus integer is int64 from the results that I was getting.

tests/java_definitions.py
richtia: This is a class.  I can move it to context.py, where it's used.

tests/functional/common.py
```diff
@@ -82,7 +83,11 @@ def substrait_function_test(
         else:
             pytest.skip("ibis expression currently undefined")
     else:
-        substrait_plan = producer.produce_substrait(sql_query, consumer)
+        if type(producer) in supported_producers:
+            substrait_plan = producer.produce_substrait(sql_query, consumer)
+        else:
+            pytest.skip(f"{type(producer).__name__} does not support the following SQL: "
+                        f"{sql_query}")
```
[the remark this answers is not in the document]
richtia: Good point.  I'll keep this in mind.

tests/java_definitions.py
richtia: Removed this in my last commit!

README.md
Thrameos: I believe current JPype supports 1.8 to 17.  There are scattered reports of problems wirh 19 but those appear mostly with jdk packaging issues rather than JPype itself.   JPype uses the oldest JNI [1.4] which should make it independent.  But includes a jar which uses Java 8 features and a accesses Java 9 via reflection.  In principle it should work with all versions of Java going forward.  Testing is limited to those versions on the azure cloud.

richtia:
@davisusanibar do you have a suggestion for how I should be pulling in the isthmus jar?  I saw you had just finished https://github.com/substrait-io/substrait-java/pull/98#

davisusanibar:
> @davisusanibar do you have a suggestion for how I should be pulling in the isthmus jar? I saw you had just finished [substrait-io/substrait-java#98](https://github.com/substrait-io/substrait-java/pull/98)

- Core: https://search.maven.org/artifact/io.substrait/core/0.1.4/jar
- Isthmus: https://search.maven.org/artifact/io.substrait/isthmus/0.1.4/jar

davisusanibar:
> > @davisusanibar do you have a suggestion for how I should be pulling in the isthmus jar? I saw you had just finished [substrait-io/substrait-java#98](https://github.com/substrait-io/substrait-java/pull/98)
> 
> * Core: https://search.maven.org/artifact/io.substrait/core/0.1.4/jar
> * Isthmus: https://search.maven.org/artifact/io.substrait/isthmus/0.1.4/jar

CLI for https://github.com/substrait-io/substrait-java/releases/tag/v0.1.4:
````
./isthmus-macOS-0.1.4  -c "CREATE TABLE Persons ( firstName VARCHAR, lastName VARCHAR, zip INT )" "SELECT lastName, firstName FROM Persons WHERE zip = 90210"

````

**Comment by davisusanibar** [comment]
> > > @davisusanibar do you have a suggestion for how I should be pulling in the isthmus jar? I saw you had just finished [substrait-io/substrait-java#98](https://github.com/substrait-io/substrait-java/pull/98)
> > 
> > 
> > 
> > * Core: https://search.maven.org/artifact/io.substrait/core/0.1.4/jar
> > * Isthmus: https://search.maven.org/artifact/io.substrait/isthmus/0.1.4/jar
> 
> CLI for https://github.com/substrait-io/substrait-java/releases/tag/v0.1.4:
> 
> ```
> ./isthmus-macOS-0.1.4  -c "CREATE TABLE Persons ( firstName VARCHAR, lastName VARCHAR, zip INT )" "SELECT lastName, firstName FROM Persons WHERE zip = 90210"
> ```

More details at: https://github.com/substrait-io/substrait-java/blob/7b195b0e561a489ec0ead7cbd8de3a961659dbf6/ci/release/README.md#how-to-use-artifacts

**Comment by aucahuasi** [comment]
> @richtia:
> @davisusanibar do you have a suggestion for how I should be pulling in the isthmus jar? I saw you had just finished https://github.com/substrait-io/substrait-java/pull/98
>> @davisusanibar:
>> Core: https://search.maven.org/artifact/io.substrait/core/0.1.4/jar
>> Isthmus: https://search.maven.org/artifact/io.substrait/isthmus/0.1.4/jar

@richtia @davisusanibar I bet we can try different approaches for integrating the java packages here.
We can create an empty `pom.xml` file (a Maven project file) just for declaring the dependencies (the java packages we want).  After that, we can run the maven command (in the same folder of the `pom.xml` file) to download those java packages and its dependencies (transitive dependencies included too), for instance: `mvn install dependency:copy-dependencies`, `mvn dependency:resolve` or even `mvn install` can do the trick.

1. We can invoke the maven command using a bash utility somewhere (perhaps together to the conda command to install all the dependencies)
2. or we can use a `setup.py` file to integrate this step as part of the python package we have in this PR (i.e. we can run custom commands through python setup tools, e.g.: `mvn ...`)

In any case, we can add all of this in a followup PR; also, we need to consider that if we want to get those java packages from the Maven repos, we may need to add maven as dependency for running the tests, in that case we should update the conda command to include maven as well.

**Comment by davisusanibar** [comment]
> > @richtia @davisusanibar I bet we can try different approaches for integrating the java packages here. We can create an empty `pom.xml` file (a Maven project file) just for declaring the dependencies (the java packages we want). After that, we can run the maven command (in the same folder of the `pom.xml` file) to download those java packages and its dependencies (transitive dependencies included too), for instance: `mvn install dependency:copy-dependencies`, `mvn dependency:resolve` or even `mvn install` can do the trick.
> > 
> > 1. We can invoke the maven command using a bash utility somewhere (perhaps together to the conda command to install all the dependencies)
> > 2. or we can use a `setup.py` file to integrate this step as part of the python package we have in this PR (i.e. we can run custom commands through python setup tools, e.g.: `mvn ...`)
> > 
> > In any case, we can add all of this in a followup PR; also, we need to consider that if we want to get those java packages from the Maven repos, we may need to add maven as dependency for running the tests, in that case we should update the conda command to include maven as well.
> 
> Thanks for the input! Since @vibhatha has a [PR out for adding a `setup.py`](https://github.com/substrait-io/consumer-testing/pull/8), maybe we should go that route. We can create a follow up PR once this and that one are both merged.

Thank you for the explanation Percy. Base on what I read we need some kind of Java fat jar that contains all java libraries dependencies. If that is needed we could add a Gradle task to also create a fat jar:

````
File: /isthmus/build.gradle.kts

import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar

plugins {
  ...
  id("com.github.johnrengelman.shadow") version "7.1.2"
}
tasks {
  named<ShadowJar>("shadowJar") {
    archiveBaseName.set("isthmus")
    manifest { attributes(mapOf("Main-Class" to "io.substrait.isthmus.PlanEntryPoint")) }
  }
}

tasks { build { dependsOn(shadowJar) } }
````

Call task with `./gradlew shadowJar`

Then use the fat jar with:
````
java -jar isthmus/build/libs/isthmus-1.0-SNAPSHOT-all.jar -c "CREATE TABLE Persons ( firstName VARCHAR, lastName VARCHAR, zip INT )" "SELECT lastName, firstName FROM Persons WHERE zip = 90210"
````

Please let me know if that is needed to we could create a PR with this feature.

**Comment by richtia** [comment]
> Thank you for the explanation Percy. Base on what I read we need some kind of Java fat jar that contains all java libraries dependencies. If that is needed we could add a Gradle task to also create a fat jar:
> 
> ```
> File: /isthmus/build.gradle.kts
> 
> import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
> 
> plugins {
>   ...
>   id("com.github.johnrengelman.shadow") version "7.1.2"
> }
> tasks {
>   named<ShadowJar>("shadowJar") {
>     archiveBaseName.set("isthmus")
>     manifest { attributes(mapOf("Main-Class" to "io.substrait.isthmus.PlanEntryPoint")) }
>   }
> }
> 
> tasks { build { dependsOn(shadowJar) } }
> ```
> 
> Call task with `./gradlew shadowJar`
> 
> Then use the fat jar with:
> 
> ```
> java -jar isthmus/build/libs/isthmus-1.0-SNAPSHOT-all.jar -c "CREATE TABLE Persons ( firstName VARCHAR, lastName VARCHAR, zip INT )" "SELECT lastName, firstName FROM Persons WHERE zip = 90210"
> ```
> 
> Please let me know if that is needed to we could create a PR with this feature.

Yes, I think this is what we need because it uses java libraries from both isthmus and substrait-java core.

**Comment by davisusanibar** [comment]
> 



> > Thank you for the explanation Percy. Base on what I read we need some kind of Java fat jar that contains all java libraries dependencies. If that is needed we could add a Gradle task to also create a fat jar:
> > ```
> > File: /isthmus/build.gradle.kts
> > 
> > import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
> > 
> > plugins {
> >   ...
> >   id("com.github.johnrengelman.shadow") version "7.1.2"
> > }
> > tasks {
> >   named<ShadowJar>("shadowJar") {
> >     archiveBaseName.set("isthmus")
> >     manifest { attributes(mapOf("Main-Class" to "io.substrait.isthmus.PlanEntryPoint")) }
> >   }
> > }
> > 
> > tasks { build { dependsOn(shadowJar) } }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Call task with `./gradlew shadowJar`
> > Then use the fat jar with:
> > ```
> > java -jar isthmus/build/libs/isthmus-1.0-SNAPSHOT-all.jar -c "CREATE TABLE Persons ( firstName VARCHAR, lastName VARCHAR, zip INT )" "SELECT lastName, firstName FROM Persons WHERE zip = 90210"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please let me know if that is needed to we could create a PR with this feature.
> 
> Yes, I think this is what we need because it uses java libraries from both isthmus and substrait-java core.

Hi @richtia. I just remember that we were talking about this feature some weeks ago, it is already implemented on main branch.

In case you need the fat Jar:

1. Create fat jar:
````
$ ../gradlew shadowJar
> Task :isthmus:shadowJar 
````

2. Go to the build directory `isthmus/build/libs`: 
````
$ ls -latr isthmus/build/libs
_ isthmus-0.1.4-sources.jar
_ isthmus-0.1.4.jar
_ isthmus-0.1.4-javadoc.jar
_ isthmus-0.1.4-all.jar // <------ fat Jar
````

3. Test Java fat JAR library:
````
$ java -jar isthmus/build/libs/isthmus-0.1.4-all.jar -c "CREATE TABLE Persons ( firstName VARCHAR, lastName VARCHAR, zip INT )" "SELECT lastName, firstName FROM Persons WHERE zip = 90210"
````

Please let us know if this is working as expected.

File: tests/context.py (hunk 1 above)
Comment (vibhatha) [reply tests/context.py]: Yes, I agree with @aucahuasi. Also we could use config files like `yml` to dynamically configure such attributes. If they are experiment bound, the user just have to do 
`consumer-testing run --config some_yaml.yaml` 

Just a thought.

**Comment by gforsyth** [comment]
I don't think we should be checking in isthmus jars to this repository -- I believe they have releases on both github releases and maven.  
Maybe a Makefile or similar to pull down the latest (or specified) release?  Otherwise this repo is going to get rather large to clone.

File: tests/functional/common.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (gforsyth) [reply tests/functional/common.py:90]: I think that skipping tests is generally not a good policy -- if a test suddenly starts passing, but it is skipped, then nothing changes.  The test isn't run, it just gets skipped.

If a test is `xfailed` and it starts passing, then you get a notification, which helps keep track of which things are working, vs. which aren't.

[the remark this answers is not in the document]
Comment (richtia) [reply tests/functional/common.py]: Good point.  I've updated it!

**Comment by richtia** [comment]
> I don't think we should be checking in isthmus jars to this repository -- I believe they have releases on both github releases and maven.
> Maybe a Makefile or similar to pull down the latest (or specified) release? Otherwise this repo is going to get rather large to clone.

Thanks for this suggestion!  I just added a Makefile and updated the README

**Review by gforsyth — COMMENTED** [review]
The code here looks good.  

I think there are some bits here that are being handled manually that should be handled in a more automated fashion.

With the hand-written tuples for which `producer` can correctly encode which query, we have the means to track regressions (`DuckDB` suddenly fails to run `logb`, e.g.) but not improvements ('Isthmus` can now encode `logb`).

This is a highly non-trivial problem, because the _outcomes_ of the tests of the producers are essentially the text fixtures for the consumers.


We've been using `pytest-snapshot` to test that Ibis produces "good" or "golden" SQL for various expressions (https://pypi.org/project/pytest-snapshot/) and I wonder if that would be of help here.

Testing producers would mean generating substrait blobs, then comparing them to known good / valid snapshots of those blobs.

Testing consumers would consist of loading the snapshots blobs and attempting to execute.

I know I'm not covering everything that needs covering in the test matrix here, but I think it would be a very good idea to start sketching out more sustainable patterns. 

Having said all of ^^^^that^^^^, I don't think that should block this PR.

I _do_ think that we should be attempting to run all producer tests on all SQL snippets, and not manually filtering them down pre-test.  If `isthmus` is going to fail one of those tests because it uses a different SQL dialect, so be it -- we can get creative in the `xfail` markers and distinguish between "tests that fail that should pass in the future" and "tests that fail that will always fail".

Alternatively, we might make use of `sqlglot` to translate string sql between dialects -- it's very good at that.

File: README.md [hunk 8]
```diff
@@ -28,14 +28,26 @@ which is located in `tests/conftest.py`.
...
 conda activate substrait_consumer_testing
 ```
+*Note: Java is used by Jpype to access the Isthmus producer.  
+JPype has been tested with Java versions from Java 1.7 to Java 13.
```
westonpace, line 35: This seems like the kind of thing that will go stale quickly.  Maybe something like "JPype should work with all versions of Java but to see details on which versions are officially supported see https://jpype.readthedocs.io/en/latest/install.html"

README.md
```diff
@@ -28,14 +28,26 @@ which is located in `tests/conftest.py`.
...
 ```commandline
 pip install -r requirements.txt
 ```
+
+Get the java dependencies needed by the Isthmus Substrait producer:
+```commandline
+git clone https://github.com/substrait-io/substrait-java.git
+export SUBSTRAIT_JAVA_HOME=$(pwd)/substrait-java
+cd consumer-testing
+make
```
westonpace, line 48: What does the error look like if the user hasn't run `make`?  Is it obvious that running `make` is what they need to do?

Similarly, if a user runs `make` and doesn't set `SUBSTRAIT_JAVA_HOME` I don't think the error will be obvious.  Can you test for the `${SUBSTRAIT_JAVA_HOME}/isthmus` directory and give a nice error about setting the environment variable if it doesn't exist?

Makefile
```diff
@@ -0,0 +1,7 @@
+all:
```
westonpace, line 1: Instead of a Makefile would it be more obvious if this was just a script named `build-and-copy-isthmus-shadow-jar.sh`?

Makefile
```diff
@@ -0,0 +1,7 @@
...
+	cd ${SUBSTRAIT_JAVA_HOME}/isthmus; ../gradlew shadowJar
+	mkdir -p jars
```
westonpace, line 3: These commands are not going to work on Windows.  This is probably ok, for now, but it should maybe be mentioned in the README somewhere.

tests/java_definitions.py
```diff
@@ -0,0 +1,24 @@
+import os
+from pathlib import Path
+
+import jpype
+
+REPO_DIR = Path(__file__).parent.parent
+calcite_jars = Path.joinpath(REPO_DIR, "jars/*")
```
westonpace, line 7: Calcite or Isthmus?

westonpace (approved):
A few thoughts but overall I'm fine with this.  I'll be out for the next two weeks.  Feel free to proceed when you and @gforsyth are satisfied.

tests/java_definitions.py
richtia: should be isthmus, thanks for the catch!

Makefile
richtia: Thanks for the suggestion, went with this approach.  The user is prompted for the absolute path of substrait-java now instead of needing to set the path first.

README.md
richtia: If the user doesn't run the script that builds/copies isthmus, but tries to run the tests still, they'll see an error like this:
```
E   TypeError: Class io.substrait.isthmus.SqlToSubstrait is not found
```

richtia:
> The code here looks good.
> 
> I think there are some bits here that are being handled manually that should be handled in a more automated fashion.
> 
> With the hand-written tuples for which `producer` can correctly encode which query, we have the means to track regressions (`DuckDB` suddenly fails to run `logb`, e.g.) but not improvements (`Isthmus` can now encode `logb`).
> 
> This is a highly non-trivial problem, because the _outcomes_ of the tests of the producers are essentially the text fixtures for the consumers.
> 
> We've been using `pytest-snapshot` to test that Ibis produces "good" or "golden" SQL for various expressions (https://pypi.org/project/pytest-snapshot/) and I wonder if that would be of help here.
> 
> Testing producers would mean generating substrait blobs, then comparing them to known good / valid snapshots of those blobs.
> 
> Testing consumers would consist of loading the snapshots blobs and attempting to execute.
> 
> I know I'm not covering everything that needs covering in the test matrix here, but I think it would be a very good idea to start sketching out more sustainable patterns.
> 
> Having said all of ^^^^that^^^^, I don't think that should block this PR.
> 
> I _do_ think that we should be attempting to run all producer tests on all SQL snippets, and not manually filtering them down pre-test. If `isthmus` is going to fail one of those tests because it uses a different SQL dialect, so be it -- we can get creative in the `xfail` markers and distinguish between "tests that fail that should pass in the future" and "tests that fail that will always fail".
> 
> Alternatively, we might make use of `sqlglot` to translate string sql between dialects -- it's very good at that.

Thanks for all this input! I definitely agree with that you're saying.  I'm thinking I can get this Isthmus producer in first and then create some github issues to make these improvements that you've suggested.

PR #8, opened 2022-11-28
feat: Python Setup integration

vibhatha:
This PR includes a modification to the existing setup and refactoring of existing helper classes and test cases. 

The changes reflects the following

- [x] Adding setup.py
- [x] Organized utils and core classes and tests to be packaged
- [x] Setup requirements clarified and documented


westonpace (commented):
These changes appear to be mostly cosmetic.  I'll defer to whatever @richtia thinks is best.  @richtia , does this look correct?

setup.py
```diff
@@ -0,0 +1,29 @@
...
+        'pyarrow',
+        'pytest',
+        'pytest-xdist',
+        'substrait-validator',
+    ],
+    dependency_links=[
+      '--extra-index-url https://pypi.fury.io/arrow-nightlies --prefer-binary --pre'
+    ],
+    package_data={
+        'tests': ['tests/integration'],
```
richtia, line 27: should tests/functional be here too?

setup.py
```diff
@@ -0,0 +1,29 @@
+from setuptools import setup, find_packages
+setup(
+    name='substrait-consumer',
```
richtia, line 3: I think this project has grown to a point where it's testing both producers and consumers.  maybe just substrait-tests? and change the folder to substrait_tests as well?  we can get feedback from others too.

setup.py
vibhatha: Ah yes, good catch.

setup.py
vibhatha: Yeah, probably we need to update the repo name too. Let’s decide a better name. +1

vibhatha: cc @jacques-n @westonpace

vibhatha:
@richtia I updated the README as well.

vibhatha:
Also, immediately after this PR we could also create a github workflow to at least test the core features of the package. WDYT @richtia?

richtia:
> Also, immediately after this PR we could also create a github workflow to at least test the core features of the package. WDYT @richtia?

I think that would be a great idea!

richtia:
@vibhatha did you run into errors trying to do the install?
```error: The 'sqlalchemy<2,>=1' distribution was not found and is required by ibis-substrait```

vibhatha:
Closing this in favor of https://github.com/substrait-io/consumer-testing/pull/12

PR #10, opened 2022-12-20
feat: test for generating substrait from adhoc sql/ibis

richtia:
test for generating substrait plans from supported producers when given a SQL query or Ibis expression.  Substrait plans can be run against supported consumers as well.

Option to save substrait plan from each producer as json output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,24 +217,24 @@ Copy your Ibis expression into `substrait_consumer/tests/adhoc/ibis_expr.py`
 Run using the following command:
 ```commandline
 cd substrait_consumer/tests/adhoc
-pytest test_adhoc_expression.py
+pytest --adhoc_producer=IsthmusProducer test_adhoc_expression.py
 ```
 
-In addition to running the plans against the consumers, you can also save the 
+In addition to running the plans against the consumers, you can save the 
 produced substrait plans by specifying the `--saveplan` option.
 ```commandline
-pytest --saveplan test_adhoc_expression.py
+pytest --saveplan True --adhoc_producer=IsthmusProducer test_adhoc_expression.py
 ```
 Plans will be saved as {producer_name}_substrait.json
 ```commandline
 ls *.json
-DuckDBProducer_substrait.json	IbisProducer_substrait.json	IsthmusProducer_substrait.json
+IsthmusProducer_substrait.json
 ```
 
-If you want to only generate plans from a specific producer or run the plans against a
-specific consumer, you can also use the `--producer` and `--consumer` options.
+If you want to run the tests from specific producer/consumer pairs, you can use 
+the `--adhoc_producer` and `--consumer` options.
 ```commandline
-pytest --producer=IsthmusProducer --consumer=AceroConsumer test_adhoc_expression.py
+pytest --adhoc_producer=IsthmusProducer --consumer=AceroConsumer test_adhoc_expression.py
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -14,13 +14,17 @@ Table of Contents
   * [Test Case Args](#Test-Case-Args)
   * [SQL Queries](#SQL-Queries)
   * [Ibis Expressions](#Ibis-Expressions)
+* [CLI Tool](#Generating-Substrait-from-Adhoc-SQL-Ibis)
+  * [How to Use](#How-to-Use)
 * [How to Add Producers](#How-to-Add-Producers)
 * [How to Add Consumers](#How-to-Add-Consumers)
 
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -198,6 +202,42 @@ IBIS_SCALAR = {
...
 
+# Generating Substrait from Adhoc SQL/Ibis
+The CLI tool for generating substrait plans from adhoc SQL queries and Ibis expression
+is located in the `substrait_consumer/tests/adhoc` directory.  The SQL queries should be 
+written using the same TPCH data used in the integration tests.  This tool will generate 
+the substrait plans for each supported producer and run that plan against all supported consumers.
+
+## How to Use
+Copy your SQL query into `substrait_consumer/tests/adhoc/query.sql`.  
+Copy your Ibis expression into `substrait_consumer/tests/adhoc/ibis_expr.py`
```
westonpace, line 213: It would probably be easier if we could specify the SQL/ibis paths from the command line?  Do I always need both an ibis and SQL file?  In other words, can I just test SQL?

README.md
```diff
@@ -198,6 +202,42 @@ IBIS_SCALAR = {
...
+
+*Note: Do not modify line 1 in ibis_expr.py
```
westonpace, line 215: I'm confused what this means.  Is `substrait_consumer/tests/adhoc/ibis_expr.py` a template I copy from?  Or is it the destination I need to save my file as?  If it is a destination then what line 1 is being discussed here?

README.md
```diff
@@ -198,6 +202,42 @@ IBIS_SCALAR = {
...
+pytest --saveplan test_adhoc_expression.py
+```
+Plans will be saved as {producer_name}_substrait.json
+```commandline
+ls *.json
+DuckDBProducer_substrait.json	IbisProducer_substrait.json	IsthmusProducer_substrait.json
+```
+
+If you want to only generate plans from a specific producer or run the plans against a
+specific consumer, you can also use the `--producer` and `--consumer` options.
```
westonpace, line 234: Does this also run the plans against the consumers?  The rest of the description only talked about producing plans.

substrait_consumer/tests/adhoc/test_adhoc_expression.py
```diff
@@ -0,0 +1,109 @@
...
+            actual_result.columns,
+            expected_result.columns,
+            message=f"Result: {actual_result.columns} "
+            f"is not equal to the expected: "
+            f"{expected_result.columns}",
+        )
+
+
+def str_to_class(classname):
+    return getattr(sys.modules[__name__], classname)
```
westonpace, line 109: Is this used?

substrait_consumer/tests/adhoc/query.sql
```diff
@@ -0,0 +1 @@
+SELECT * FROM lineitem
```
vibhatha, line 1: maybe newline

README.md
richtia: Yes it does.  I'll update the description.

substrait_consumer/tests/adhoc/test_adhoc_expression.py
richtia: Thanks for the catch.  I'll remove this.

README.md
richtia: It's the location where the ibis expression should be written.  I'm expecting a specific name for the ibis expression as well as all the TPCH tables to be passed there, that's what the note is supposed to be for.  I can make it more descriptive.

README.md
richtia: That makes sense.  By default it's going through both right now because i'm using some existing fixtures from the test framework.  I'll create a separate fixture and force the user to specify which producers they want to test out instead.  We'll only check for the ibis file if they specify `IbisProducer`, otherwise we'll check for the SQL file.

substrait_consumer/conftest.py
```diff
@@ -41,6 +41,24 @@ def pytest_addoption(parser):
         help="A comma separated list of producers to run against.",
         choices=[x.__name__ for x in PRODUCERS]
     )
+    parser.addoption(
+        "--adhoc_producer",
+        action="store",
+        default=",".join([x.__name__ for x in []]),
```
westonpace, line 47: Is this default a fancy way of saying "empty list" or am I missing something?

westonpace (approved):

substrait_consumer/conftest.py
richtia: Oops, this can actually just be an empty string! I'll update it.

PR #11, opened 2022-12-24
Add velox consumer

richtia:


ingomueller-net:
This PR seems to have gone out of sync with `main` and has been inactive for two years. Shall we close it, @richtia?

PR #14, opened 2023-02-11
Use pytest-snapshot package for better validation

richtia:
This PR separates producers tests from consumer tests.

1. Producer tests will validate against a known `golden substrait plan`.
2. Consumer will consume the `golden substrait plan` and generate results, which are also saved as a set of `golden results`

Separating these test allows developers of producers or consumers to more easily validate results or debug issues.  This will also make it easier for future reporting.

This PR saves the `golden substrait plan` in json format at the moment.  The Isthmus producer currently isn't able to generate the binary format that's consumable by consumers, so I've stuck with json for consistency.  Having it in json format also makes it easier to see debug when plans end up changing.


README.md
```diff
@@ -151,9 +151,20 @@ used.
...
-running the substrait plans on different consumers.  The results of running the substrait plans
-are then compared to an expected result, which is determined by running the original query
-against a trusted engine (currently we use DuckDB).
+running the substrait plans on different consumers.  
+
+The producer tests take the SQL/Ibis query and generate a substrait plan in json format.  
+These plans are saved as snapshots to be used later on for verification as well as an input to 
+the consumer tests.
+
+The consumer tests read the saved substrait plan snapshots and generate results.  These
```
vibhatha, line 160:
```suggestion
The consumer tests read the saved substrait plan snapshots and generate results. These
```

vibhatha:
@richtia We should probably avoid committing submodule diff to this repo. We shoud probably update the submodule, `git submodule update --init`

substrait_consumer/consumers.py
```diff
@@ -26,21 +26,22 @@ def __init__(self, db_connection=None):
         self.db_connection.execute("INSTALL substrait")
         self.db_connection.execute("LOAD substrait")
 
-    def setup(self, db_connection, file_names: Iterable[str]):
+    def setup(self, db_connection, created_tables, file_names: Iterable[str]):
         self.db_connection = db_connection
+        self.load_tables_from_parquet(created_tables, file_names)
 
-    def run_substrait_query(self, substrait_query: bytes) -> pa.Table:
```
vibhatha, line 32: Why are we changing this interface? Can we keep this as it is an may be in the caller we convert the `json` string to bytes?

substrait_consumer/functional/common.py
```diff
@@ -34,20 +35,20 @@ def check_subtrait_function_names(
     ), f"Error in function: {expected_function_name}.  Does not appear in {functions_list}."
 
 
-def substrait_function_test(
+def substrait_producer_function_test(
+    test_name,
+    snapshot,
```
vibhatha, line 40: should we add docstrings for these two parameters along with type hints?

substrait_consumer/functional/common.py
```diff
@@ -77,28 +75,70 @@ def substrait_function_test(
...
+
+    function_group = test_name.split(":")[0]
+    function_name = test_name.split(":")[1]
+    snapshot.snapshot_dir = f"{function_group}/{type(producer).__name__}"
+    snapshot.assert_match(str(substrait_plan), f"{function_name}_plan.json")
+
+
+def substrait_consumer_function_test(
+    test_name,
+    snapshot,
```
vibhatha, line 98: same goes here, add docstrings and type hints?

vibhatha:
@richtia @westonpace @gforsyth since we have a lot of text files, should we use a separate repo or blobstore to keep the files and use it as a submodule in this repo. I am suggesting something [similar to what's done in Apache Arrow](https://github.com/apache/arrow-testing/tree/d2c73bf78246331d8e58b6f11aa8aa199cbb5929).

vibhatha:
@richtia Thanks for working on this. I added a few initial review comments.

substrait_consumer/tests/functional/extension_functions/arithmetic_decimal_snapshots/DuckDBConsumer/DuckDBProducer/divide_result.txt
```diff
@@ -0,0 +1,8 @@
+pyarrow.Table
+l_tax: decimal128(15, 2)
+l_extendedprice: decimal128(15, 2)
+DIVIDE_KEY: double
+----
+l_tax: [[0.02,0.06,0.02,0.06,0.04,...,0.06,0.05,0.04,0.00,0.00]]
+l_extendedprice: [[24386.67,58958.28,10210.96,31197.88,31329.60,...,36739.17,16112.40,57084.96,8859.36,1828.91]]
+DIVIDE_KEY: [[8.201201722088339e-7,0.0000010176687650996603,0.000001958679693192413,0.0000019232076025678665,0.0000012767478678310608,...,0.000001633134335914502,0.0000031032000198604804,7.007099593307939e-7,0,0]]
```
gforsyth, line 8: I think you'll want to add a `round` or a `truncate` or something to any floating point results, because otherwise this is going to be a flaky indicator of changes.

gforsyth: or cast to decimal?

substrait_consumer/functional/common.py
```diff
@@ -1,10 +1,11 @@
+from pathlib import Path
 from typing import Callable, Iterable
 
 import pytest
 from duckdb import DuckDBPyConnection
 from ibis.expr.types.relations import Table
 
-from substrait_consumer.verification import verify_equals
+SNAPSHOT_DIR = Path(__file__).parent.parent / f"tests/functional/extension_functions"
```
gforsyth, line 8: This doesn't need to be an `f-string` -- I think `flake8` should catch that.

And it probably doesn't matter, but if there's a thought that this might be run on Windows (ever) then you'll want to divide each part of the path individually so that the path separators are compatible, e.g.

```
SNAPSHOT_DIR = Path(__file__).parent.parent / "tests" / "functional" / "extension_functions"
```

gforsyth (commented):
I would suggest adding all of the generated files (you can use a glob) to the `.gitattributes` file and mark them as `linguist-generated` so that (generally) the changes are collapsed during github reviews (just to keep things manageable for reviewers):
https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

westonpace (commented):
A few top-level questions:

Each producer has a snapshot set of plans.  This makes sense.
Each consumer has a snapshot set of results for each producer.  Why is it "for each producer"?  Wouldn't we expect the query results to be the same regardless of producer?

What do the workflows look like:

New Producer:
 If we are adding a new producer we need to create new snapshots for all of the plans but we don't care about the results snapshots right? (I suppose this depends on the above answer)

New Consumer:
  If we are adding a new consumer we need to create new snapshots for all of the results but we can use the existing snapshots of the plans right?

Producer mismatch:
  If a producer test fails that means it generated a different plan.  If it turns out that plan was the correct plan is it possible to just update that one snapshot?  Or maybe the change is systemic and we need to update all the snapshots so both scenarios seem likely.

Consumer mismatch:
  Same as above, maybe we realize that we were doing function `xyz` incorrectly and so we fix it.  Can we just replace the one snapshot?

It's not clear to me, and maybe it's time to write a tutorial or "getting started" that covers what the whole workflow will be like for someone that is creating a producer or consumer.

For example, pretend I am creating a consumer:

 * Do I need to get my consumer added to this repo?  Or can I clone this repo and use it internally to do my own testing with the supported producers?
 * What does that process look like?  What commands will I run?  How will I integrate this with CI?

richtia:
> A few top-level questions:
> 
> Each producer has a snapshot set of plans. This makes sense. Each consumer has a snapshot set of results for each producer. Why is it "for each producer"? Wouldn't we expect the query results to be the same regardless of producer?
> 

That's a good point.  My original plan was to pre-generate the snapshots just using DuckDB, but I figured that would make updating the snapshots require a few more steps.  That probably makes the most sense though.  I'll go with that route and add the instructions in the README.

> What do the workflows look like:
> 
> New Producer: If we are adding a new producer we need to create new snapshots for all of the plans but we don't care about the results snapshots right? (I suppose this depends on the above answer)
> 

Yep.


> New Consumer: If we are adding a new consumer we need to create new snapshots for all of the results but we can use the existing snapshots of the plans right?
> 

Yep.

> Producer mismatch: If a producer test fails that means it generated a different plan. If it turns out that plan was the correct plan is it possible to just update that one snapshot? Or maybe the change is systemic and we need to update all the snapshots so both scenarios seem likely.
> 

Yeah, we should be able to update just that one snapshot.  Updating snapshots is made pretty simple by the pytest-snapshot plugin, so I don't think this is too much of an issue.

> Consumer mismatch: Same as above, maybe we realize that we were doing function `xyz` incorrectly and so we fix it. Can we just replace the one snapshot?
> 

Yep.

> It's not clear to me, and maybe it's time to write a tutorial or "getting started" that covers what the whole workflow will be like for someone that is creating a producer or consumer.
> 
> For example, pretend I am creating a consumer:
> 
> * Do I need to get my consumer added to this repo?  Or can I clone this repo and use it internally to do my own testing with the supported producers?
> * What does that process look like?  What commands will I run?  How will I integrate this with CI?

I'll add a tutorial/getting started to this PR.  It's definitely possible to clone this repo and use it internally.  How to add a consumer is already in the README and is relatively simple, but I need to update the process for updating the snapshots too.  I'll open a separate PR for the CI part.

substrait_consumer/consumers.py
richtia: I changed it to be consistent since all the producers can create the json plan, but not all producers can produce the byte representation (isthmus cannot)

substrait_consumer/functional/common.py
```diff
@@ -77,28 +123,76 @@ def substrait_function_test(
...
+            Ibis expression.
+        producer:
+            Substrait producer class.
+        consumer:
+            Substrait consumer class.
+    """
+    consumer.setup(db_con, created_tables, file_names)
+
+    function_group = test_name.split(":")[0]
+    function_name = test_name.split(":")[1]
```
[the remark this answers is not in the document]
richtia: I like this better too :)

gforsyth (approved):
this looks good to me.

PR #17, opened 2023-04-03
test: add tests for isthmus arithmetic functions

richtia:
PR to bump substrait-java submodule version and add support for more isthmus producer tests


richtia:
> Seems reasonable to me. The cast to double is needed because Isthmus doesn't do any kind of implicit casting?

Yep, that's right.  Isthmus just errors out without the cast :(

richtia:
Need to bump the substrait-java version to get the latest isthmus changes with round functions.  This is needed to generate the isthmus plans in this PR

PR #21, opened 2023-10-11
feat: add datafusion consumer

richtia:
PR to add the datafusion consumer to integration test suite.

This PR does not yet include adding additional tests for the datafusion consumer.

### DIFF
--- a/substrait_consumer/consumers.py
+++ b/substrait_consumer/consumers.py
@@ -8,6 +8,8 @@ import duckdb
 import pyarrow as pa
 import pyarrow.parquet as pq
 import pyarrow.substrait as substrait
+from datafusion import SessionContext
+from datafusion import substrait as ds
 
 from substrait_consumer.common import SubstraitUtils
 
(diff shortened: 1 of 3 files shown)


substrait_consumer/consumers.py
```diff
@@ -146,3 +148,82 @@ def run_substrait_query(self, substrait_query: str) -> pa.Table:
...
+                table_name = table_name.translate(
+                    str.maketrans("", "", string.punctuation)
+                )
+                if f"{self.__class__.__name__}{table_name}" not in created_tables:
+                    created_tables.add(f"{self.__class__.__name__}{table_name}")
+                    self.ctx.register_parquet(f"{table_name}", file_path)
+        else:
+            tables = pa.RecordBatch.from_arrays(
+                [
+                    pa.array([1, 2, 3, -4, 5, -6, 7, 8, 9, None]),
```
EpsilonPrime, line 175: If these are the same as other consumers perhaps we could extract this out?

EpsilonPrime (approved):

substrait_consumer/consumers.py
richtia: Good point.  Done!

EpsilonPrime (approved):

PR #22, opened 2023-10-18
feat: initial commit to add datafusion consumer tests

richtia:
This PR includes the following changes:
1.  Update the datafusion consumer to accept json output plans
2.  Update to snapshot results to be only the results data


substrait_consumer/tests/functional/extension_functions/boolean_snapshots/function_test_results/xor_result.txt
```diff
@@ -1,8 +1,10 @@
-pyarrow.Table
-a: int64
-b: int64
-xor_a_b: int64
-----
-a: [[1,2,3,-4,5,-6,7,8,9,null]]
-b: [[1,1,1,1,1,2,2,2,2,2]]
-xor_a_b: [[0,3,2,-3,4,-8,5,10,11,null]]
\ No newline at end of file
+1
```
EpsilonPrime, line 1: Do we want the xor_a_b value here?

richtia: oh yeah, we do.  thanks for the catch.

substrait_consumer/functional/common.py
```diff
@@ -200,8 +202,11 @@ def substrait_consumer_function_test(
         actual_result = consumer.run_substrait_query(substrait_plan)
         actual_result = actual_result.rename_columns(
             list(map(str.lower, actual_result.column_names))
-        ).columns[0].data
-        str_result = '\n'.join(map(str, actual_result))
+        ).columns
+        result_list = []
+        for column in actual_result:
```
EpsilonPrime, line 207: Could we add a blank line between columns?  That'll make it possible to tell the unlikely difference between an item shifting from the end of one column to the front of the next.

EpsilonPrime (approved):

substrait_consumer/conftest.py
```diff
@@ -63,7 +63,7 @@ def saveplan(request):
 
 
 PRODUCERS = [DuckDBProducer, IbisProducer, IsthmusProducer]
-CONSUMERS = [AceroConsumer, DuckDBConsumer]
```
EpsilonPrime, line 66: Does A cero not work?

richtia: It sort of works....but most of these tests will fail because Acero doesn't support projection in the `ReadRel`

https://github.com/apache/arrow/issues/14965

I don't think there is a plan to really continue work on the Acero Consumer either.

EpsilonPrime (approved):
Still looks good.  I haven't looked at all of the generated changes (given the sheer volume of them) but I've spot checked many of them.

richtia:
> Still looks good. I haven't looked at all of the generated changes (given the sheer volume of them) but I've spot checked many of them.

Most of the substrait plan json revisions are due to updates in the producers (the others are from me adding some limits to the output).  Eventually we should set something up so that those plans can automatically get updated when the substrait producers are updated.

PR #23, opened 2023-10-25
feat: add support for datafusion producer

richtia:
feat: add support for datafusion producer

### DIFF
--- a/substrait_consumer/producers.py
+++ b/substrait_consumer/producers.py
@@ -4,11 +4,17 @@ from pathlib import Path
 from typing import Iterable
 
 import duckdb
+import pyarrow as pa
 import pytest
+import substrait.gen.proto.plan_pb2 as plan_pb2
+from datafusion import SessionContext
+from datafusion import substrait as ss
 from google.protobuf import json_format
+from google.protobuf.json_format import MessageToJson
 from ibis_substrait.compiler.core import SubstraitCompiler
 
 from substrait_consumer.common import SubstraitUtils
+from substrait_consumer.consumers import COLUMN_A, COLUMN_B, COLUMN_C, COLUMN_D
 from substrait_consumer.context import get_schema, produce_isthmus_substrait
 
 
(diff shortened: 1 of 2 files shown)


substrait_consumer/producers.py
```diff
@@ -138,6 +144,83 @@ def name(self):
         return "IsthmusProducer"
 
 
+class DataFusionProducer:
```
vibhatha, line 147: @richtia should we add short description about this producer in docstrings?

substrait_consumer/producers.py
```diff
@@ -138,6 +144,83 @@ def name(self):
...
+    def __init__(self, db_connection=None):
+        self.ctx = SessionContext()
```
vibhatha, line 149: nit: maybe

```suggestion
        self._ctx = SessionContext()
```

substrait_consumer/producers.py
```diff
@@ -138,6 +144,83 @@ def name(self):
...
+        if db_connection is not None:
+            self.db_connection = db_connection
```
vibhatha, line 151: if `self._` approach is adopted we may need reflect that change to similar attributes.

substrait_consumer/producers.py
```diff
@@ -138,6 +144,83 @@ def name(self):
...
+        else:
+            self.db_connection = db_connection
+
+    def set_db_connection(self, db_connection):
+        self.db_connection = db_connection
+
+    def produce_substrait(self, sql_query: str, ibis_expr: str = None) -> str:
```
vibhatha, line 158: @richtia just a thought, it seems like `produce_substrait` is a behavior which is common to all the producers. I would suggest to create a base class and extend from that. But of course it is out of scope for this PR.

substrait_consumer/producers.py
```diff
@@ -138,6 +144,83 @@ def name(self):
...
+
+        substrait_plan = ss.substrait.serde.serialize_to_plan(sql_query, self.ctx)
+        substrait_plan_bytes = substrait_plan.encode()
+        substrait_proto.ParseFromString(substrait_plan_bytes)
+
+        return MessageToJson(substrait_proto)
+
+    def register_tables(self, created_tables, file_names):
+        """
+        Register tables to the datafusion session context
```
vibhatha, line 178: nit: 

```suggestion
        Register tables to the datafusion session context.
```

vibhatha:
@richtia I left a few comments, but overall the PR looks good to me. :slightly_smiling_face:

substrait_consumer/producers.py
richtia: Good point.  Looks like I missed adding them to all the producers.  I'll put something simple, like what we have in consumers.py

substrait_consumer/producers.py
richtia: Agreed.  initially it was like that, but i didn't make too much sense because there was only 1-2 producers.  I'll create a backlog issues.

substrait_consumer/producers.py
```diff
@@ -138,6 +153,86 @@ def name(self):
...
+
+    def produce_substrait(self, sql_query: str, ibis_expr: str = None) -> str:
+        """
+        Produce the DataFusion substrait plan using the given SQL query.
+
+        Parameters:
+            sql_query:
+                SQL query.
+        Returns:
+            Substrait query plan in json format.
```
EpsilonPrime, line 178: Raises DecodeError if the plan cannot be parsed.

substrait_consumer/producers.py
```diff
@@ -138,6 +153,86 @@ def name(self):
...
+        self._ctx = SessionContext()
+        if db_connection is not None:
+            self._db_connection = db_connection
+        else:
+            self._db_connection = db_connection
+
+    def set_db_connection(self, db_connection):
+        self._db_connection = db_connection
+
+    def produce_substrait(self, sql_query: str, ibis_expr: str = None) -> str:
```
EpsilonPrime, line 170: ibis_expr is not documented below

EpsilonPrime (approved):

substrait_consumer/producers.py
richtia: ibis_expr is only used with the IbisProducer

vibhatha (approved):

EpsilonPrime (approved):

PR #28, opened 2023-11-15
feat: add duckdb tests ci

richtia:
This PR adds the CI to run DuckDB producer and consumer tests


richtia:
Most of the duckdb consumer failures are due to this issue: https://github.com/duckdb/substrait/issues/65

Both ibis and isthmus use the full compound function name, which includes the types

substrait_consumer/context.py
```diff
@@ -2,10 +2,8 @@
 
 import jpype.imports
 
-import substrait_consumer.java_definitions as java
```
EpsilonPrime, line 5: Seems odd to move jmoorts down, is it because you want to be able to only include them conditionally?

EpsilonPrime (approved):

substrait_consumer/context.py
richtia: yeah, that's exactly it...otherwise we need to install java on some of the docker containers that don't need it.

PR #30, opened 2023-11-16
feat: Add CI for Ibis substrait producer tests

richtia:
* add CI for Ibis substrait producer tests
* update ibis substrait plans
* pin the substrait version so that we don't need to update the ibis plans with every new substrait release

### DIFF
--- a/ci/docker/ibis_producer.Dockerfile
+++ b/ci/docker/ibis_producer.Dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:22.04
 ENV PYTHONUNBUFFERED=1
 RUN apt-get update && apt-get install -y python3.10 && ln -sf python3 /usr/bin/python
 RUN apt install -y pip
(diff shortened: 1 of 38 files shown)


richtia:
some ibis-substrait updates are causing any expressions that use rounding to fail.  looking into this.

ci/docker/ibis_producer.Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:22.04
+
+ENV PYTHONUNBUFFERED=1
+RUN apt-get update && apt-get install -y python3.10 && ln -sf python3 /usr/bin/python
+RUN apt install -y pip
+RUN pip install --upgrade pip setuptools pytest pytest-snapshot substrait==0.8.0 pyarrow protobuf duckdb filelock datafusion ibis_substrait ibis_framework=7.0.0 JPype1
```
EpsilonPrime, line 6: Substrait last released at 0.38.0 -- should we use that instead?

richtia:
> some ibis-substrait updates are causing any expressions that use rounding to fail. looking into this.

The issues I were running into are being fixed here: https://github.com/ibis-project/ibis-substrait/pull/823

EpsilonPrime (approved):

PR #38, opened 2023-12-06
refactor: decouple producers.py and consumers.py

richtia:
This PR creates abstract base classes for consumers and producers.
The single file being used for each was getting too large as we kept adding
more producers and consumers.


substrait_consumer/consumers/acero_consumer.py
```diff
@@ -0,0 +1,65 @@
...
+import pyarrow.substrait as substrait
+
+from substrait_consumer.common import SubstraitUtils
+
+from .consumer import COLUMN_A, COLUMN_B, COLUMN_C, COLUMN_D, Consumer
+
+
+class AceroConsumer(Consumer):
+    """
+    Adapts the Acero Substrait consumer to the test framework.
```
EpsilonPrime, line 18: The words "test framework" feel off.  You are probing/validating behavior by running queries.  When I hear the word test I am thinking that you're testing this code which isn't the case.  Not worth changing at this point, however.

substrait_consumer/consumers/acero_consumer.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self):
+        self.tables = {}
+        self.table_provider = lambda names, schema: self.tables[names[0].lower()]
+
+    def setup(self, db_connection, created_tables, file_names: Iterable[str]):
+        if len(file_names) > 0:
+            parquet_file_paths = SubstraitUtils.get_full_path(file_names)
+            for file_name, file_path in zip(file_names, parquet_file_paths):
+                table_name = Path(file_name).stem
```
EpsilonPrime, line 29: Eventually should the logic to create a table be another abstract method on the base class?

substrait_consumer/consumers/datafusion_consumer.py
```diff
@@ -0,0 +1,73 @@
...
+        Run the substrait plan against Datafusion.
+
+        Parameters:
+            substrait_query:
+                A substrait plan in bytes.
+
+        Returns:
+            A pyarrow table resulting from running the substrait query plan.
+        """
+        substrait_json = json.loads(substrait_query)
```
EpsilonPrime, line 63: This is the only class using it, but this could be a utility function.

EpsilonPrime (approved):

substrait_consumer/consumers/acero_consumer.py
richtia: Makes sense...i'll create an issue to track this.

PR #39, opened 2023-12-11
Fix: fix test results path in streamlit app python script

richtia:
Test results path needs to be a url in order for streamlit to be able to read it
since streamlit is a third party app.

### DIFF
--- a/app/substrait_consumer_function_compatibility.py
+++ b/app/substrait_consumer_function_compatibility.py
@@ -10,7 +10,9 @@ st.set_page_config(layout="wide")
 
 def support_matrix_df():
     return (
-        ibis.read_csv("./consumer_results.csv")
+        ibis.read_csv(
+            "https://raw.githubusercontent.com/substrait-io/consumer-testing/main/app/consumer_results.csv"
+        )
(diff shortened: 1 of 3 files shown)


app/substrait_consumer_function_compatibility.py
```diff
@@ -10,7 +10,9 @@
 
 def support_matrix_df():
     return (
-        ibis.read_csv("./consumer_results.csv")
+        ibis.read_csv(
+            "https://raw.githubusercontent.com/substrait-io/consumer-testing/main/app/consumer_results.csv"
```
EpsilonPrime, line 14: Does this need to be an externally accessible URL or does it merely need to be in URI form?  I'm surprised that the website does calculations -- I would have expected it to work off precomputed results.

EpsilonPrime (approved):
We should get this in to fix the website.

app/substrait_consumer_function_compatibility.py
richtia: I think it needs to be an externally accessible URL.  Maybe at least for their free community cloud.  I think they just access this script and try to run it from their cloud.  I'll get this in and try some other things.

PR #43, opened 2024-01-02
fix: add aceroconsumer back to conftest

richtia:
Adding back the AceroConsumer to conftest so it can be tested with


vibhatha (approved):
@richtia any specific reason why this was removed before?

richtia:
> @richtia any specific reason why this was removed before?

Nothing other than Acero development stopping.

vibhatha:
@richtia should we worry about these failed checks?

vibhatha:
And @richtia Thanks for working on this very swiftly.

PR #92, opened 2024-08-14
fix: add pytest mark to datafusion tests

richtia:


substrait_consumer/tests/functional/relations/test_sort_relation.py
```diff
@@ -60,6 +73,7 @@ def test_producer_sort_relations(
 
     @custom_parametrization(SORT_RELATION_TESTS)
     @pytest.mark.consume_substrait_snapshot
+    @pytest.mark.usefixtures('mark_consumer_tests_as_xfail')
```
EpsilonPrime, line 76: You can avoid adding the annotation here if you define the marking fixture as autouse:

https://github.com/voltrondata/spark-substrait-gateway/blob/main/src/gateway/tests/test_dataframe_api.py#L75

richtia: i only want the fixture to be marked against the consumer tests

EpsilonPrime (approved):

PR #103, opened 2024-09-06
feat: add aggregate relation tests

richtia:


substrait_consumer/tests/functional/relations/aggregate_relation_snapshots/DuckDBProducer/aggregate_with_group_by_cube_plan.json
```diff
@@ -0,0 +1,290 @@
...
+                                }
+                              }
+                            }
+                          ]
+                        }
+                      }
+                    ]
+                  }
+                },
+                "expressions": [
```
EpsilonPrime, line 216: These project after aggregate expressions in Datafusion and DuckDB are technically correct but are weird.  Also these add new fields (but don't restrict the output) which mean that the total number of names is going to be wrong.

substrait_consumer/tests/functional/relations/aggregate_relation_snapshots/IsthmusProducer/aggregate_with_group_by_cube_plan.json
```diff
@@ -0,0 +1,300 @@
...
+        "sort": {
+          "common": {
+            "direct": {
+            }
+          },
+          "input": {
+            "project": {
+              "common": {
+                "emit": {
+                  "outputMapping": [3, 4, 5]
```
EpsilonPrime, line 24: This project is hiding whether or not the aggregate outputs an additional column.

substrait_consumer/tests/functional/relations/aggregate_relation_snapshots/DuckDBProducer/aggregate_with_group_by_cube_plan.json
richtia: do you think that's something that the duckdb/datafusion folks should "fix"?

substrait_consumer/tests/functional/relations/aggregate_relation_snapshots/IsthmusProducer/aggregate_with_group_by_cube_plan.json
richtia: could you explain a little more? how is it doing that?

EpsilonPrime: Hmm, so the output fields were 0, 1, 2 and they are being reexposed here as three fields.   If there was an extra column (needed if there are two or more grouping sets) then the columns should be 4, 5, 6.  It's not clear if they're treating the extra column as the first remapped column.  I suppose the correctness tests and types could validate that.

substrait_consumer/tests/functional/relations/aggregate_relation_snapshots/DuckDBProducer/aggregate_with_group_by_cube_plan.json
EpsilonPrime: It's something that should be fixed.

EpsilonPrime (approved):
For the future:  the producer tests are not satisfying to me as a test methodology.  Version numbers should only really matter if we're testing at a particular Substrait version.  The validator is testing plan generation validity and the tests validate correctness (at least for the platforms we try it against).  If the validator is doing its job then we shouldn't need to store the generated plans.  For correctness I'd argue that a Producer has to run its plan against its own system and at least one alternative successfully for it to be correct.

richtia:
> For the future: the producer tests are not satisfying to me as a test methodology. Version numbers should only really matter if we're testing at a particular Substrait version. The validator is testing plan generation validity and the tests validate correctness (at least for the platforms we try it against). If the validator is doing its job then we shouldn't need to store the generated plans. For correctness I'd argue that a Producer has to run its plan against its own system and at least one alternative successfully for it to be correct.

Yeah, that was the original idea for this testing framework.  Hopefully we can get to that point eventually where producers/consumers are mature enough to do that.

PR #105, opened 2024-09-11
fix: update isthmus tpch substrait plans

richtia:


substrait_consumer/tests/integration/queries/tpch_substrait_plans/query_2_plan.json
EpsilonPrime, line 1: We could name the plans with a leading zero so they sort properly.

EpsilonPrime (approved):

PR #109, opened 2024-09-19
fix: update to use latest duckdb and modify groupby relation tests

richtia:
Mostly updates to duckdb produced substrait plans.

Made the SQL for groupby relations (groupby, rollup, cube) more similar
so it's easier to understand the differences when comparing substrait plans


EpsilonPrime (approved):
Over half of the changes were the version marker changing.  It would be nice to exclude that from the difference so that we don't have to review all of the files every time a backend updates the Substrait dependency.

PR #112, opened 2024-09-24
fix: update datafusion to latest

richtia:
* Update Datafusion to latest
* Add a few new substrait plans that can now be generated
* Changes `ends_with` due to different function names


ci/docker/datafusion_consumer_functions.Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM ubuntu:22.04
 ENV PYTHONUNBUFFERED=1
 RUN apt-get update && apt-get install -y python3.10 && ln -sf python3 /usr/bin/python
 RUN apt install -y pip
-RUN pip install --upgrade pip setuptools pytest pytest-snapshot substrait pyarrow protobuf duckdb filelock datafusion==40.1.0 ibis_substrait JPype1 substrait-validator
+RUN pip install --upgrade pip setuptools pytest pytest-snapshot substrait pyarrow protobuf duckdb filelock datafusion==41.0.0 ibis_substrait JPype1 substrait-validator
```
EpsilonPrime, line 6: We're referring to this same version in at least a dozen places.  Perhaps we could put this in a file of constant version dependencies so we don't miss any locations?

substrait_consumer/functional/queries/sql/string_functions_sql.py
```diff
@@ -40,14 +40,22 @@
         """,
         [DataFusionProducer],
     ),
-    "ends_with": (
+    "ends_with_duckdb": (
```
EpsilonPrime, line 43: A comment here noting that providing the query in DuckDB SQL is needed to generate Substrait using the intended function could be helpful.

substrait_consumer/tests/functional/extension_functions/arithmetic_snapshots/DataFusionProducer/atan2_plan.json
```diff
@@ -3,14 +3,14 @@
     {
       "extensionFunction": {
         "extensionUriReference": 4294967295,
-        "name": "atan2"
+        "functionAnchor": 1,
```
EpsilonPrime, line 6: The output order of the extension functions flipped but the values are the same.  Normalization of the URI and function sections is probably useful to avoid diffs at runtime.  Here we probably want to normalize so that we don't have to resubmit the results if the order changes in a subsequent run.

substrait_consumer/tests/functional/extension_functions/string_snapshots/DataFusionProducer/string_agg_plan.json
```diff
@@ -0,0 +1,118 @@
+{
+  "extensions": [
+    {
+      "extensionFunction": {
+        "extensionUriReference": 4294967295,
```
EpsilonPrime, line 5: Is referring to a non-existent reference legal?

Issue #122, opened 2024-10-31
Table creation is broken


ingomueller-net:
The current table creation mechanism seems broken. I run the tests from the root directory of the repository as follows:

```
unbuffer pytest -v substrait_consumer/tests/ | tee pytest.out
```

Then I get several errors either due to tables not being found or tables being created multiple times:

```
$ grep -oE '(KeyError: '.*')|(Catalog Error: Table with.*!)' pytest.out | sort | uniq -c
     10 Catalog Error: Table with name "customer" already exists!
     12 Catalog Error: Table with name "lineitem" already exists!
     36 Catalog Error: Table with name lineitem does not exist!
      6 Catalog Error: Table with name nation does not exist!
      4 Catalog Error: Table with name "orders" already exists!
      6 Catalog Error: Table with name "part" already exists!
      4 Catalog Error: Table with name "partsupp" already exists!
    102 Catalog Error: Table with name partsupp does not exist!
      8 Catalog Error: Table with name "supplier" already exists!
      3 KeyError: 'customer'
      8 KeyError: 'customer_small'
      9 KeyError: 'extensions'
      9 KeyError: 'extensions'
      2 KeyError: 'lineitem'
     11 KeyError: 'lineitem_small'
      2 KeyError: 'orders'
      8 KeyError: 'orders_small'
      2 KeyError: 'partsupp_small'
      2 KeyError: 'region_small'
```

I have tried to run a couple of the failing tests in isolation and they all either passed or failed with a different error. Digging a bit deeper suggested that the protocol of storing the names of tables that consumers or producers have created in `created_tables` is broken. Some places prepend their class name to the front of the table name but some don't, so they either think that a table name already exists even though it doesn't (because some other place create a table with the same name but in a different engine) or they think that a table does not exist even though it does (because some other place already created a table with that name but registered it with a different prefix).

ingomueller-net:
Resolved by #123.

PR #123, opened 2024-10-31
fix: remove tracking of existing tables in `created_tables`

ingomueller-net:
As raised in #122, the current usage of the table creation mechanism is broken because of wrong book keeping of which tables have been created. This has previously been implemented with a set `created_tables` than was handed around different places that used or created tables. The wrong book keeping lead to tables not existing when test queries would run or tables already existing when they should be created.

This PR gets rid of the `created_tables` set all together. Instead, those places responsible for creating tables is expected to do its own book keeping in order to find out which tables need to be created (and all already do that book keeping). This PR implements this logic currently in such a way that existing tables with the same name as those that should be created are first dropped and then recreated. It is conceivable to be more optimistic and not do anything if the table already exists; however, that mechanism breaks if two different tables are loaded with the same name and is thus more brittle.

The PR also fixes `test_substrait_function_names.py`, which used an additional DuckDB connection to prepare the Substrait plan used for test but did not load the tables used in the plans. The PR thus adds the logic to load those tables.

I have compared the list of failing tests before and after this change: no test newly fails and 15 previously failing tests now pass. However, there is a much larger number of tests affected; they now just fail because the corresponding system under test fails rather than because the test harness is broken.

This resolves #112.

### DIFF
--- a/substrait_consumer/consumers/acero_consumer.py
+++ b/substrait_consumer/consumers/acero_consumer.py
@@ -22,17 +22,12 @@ class AceroConsumer(Consumer):
         self.tables = {}
         self.table_provider = lambda names, schema: self.tables[names[0].lower()]
 
-    def setup(self, db_connection, created_tables, file_names: Iterable[str]):
+    def setup(self, db_connection, file_names: Iterable[str]):
         if len(file_names) > 0:
             parquet_file_paths = SubstraitUtils.get_full_path(file_names)
             for file_name, file_path in zip(file_names, parquet_file_paths):
                 table_name = Path(file_name).stem
-                table_name = table_name.translate(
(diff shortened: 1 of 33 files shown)


EpsilonPrime (approved):

EpsilonPrime:
Thanks for the fix!

I'm not sure why this was added in the first place but if it's not currently helping taking it out seems reasonable to me.  Might want to wait to see if @richtia who touched the consumers last has any thoughts.

richtia:
Thanks for fixing this! The bookkeeping worked originally when there weren't as many producers/consumers. We've had to handle it differently since. I think this issues wasn't showing up in the CI as often because the CI only species a single producer in each run.

PR #124, opened 2024-11-01
feat: make producer and consumer CLI arguments nicer

ingomueller-net:
This PR changes the string that is given to select consumers and producers from being the respective class name to a shorter string. For example `--producer=DuckDBProdcuer` becomes `--producer=duckdb`.


ingomueller-net:
I have run the commands in `SUPPORT_MATRIX.md` and they seem to work but I haven't tested the connection all the way to the web app, so there might be some more places that need to be adapted. With I bit of pointers, I might be able to test and change those as well.

richtia:
> I have run the commands in `SUPPORT_MATRIX.md` and they seem to work but I haven't tested the connection all the way to the web app, so there might be some more places that need to be adapted. With I bit of pointers, I might be able to test and change those as well.

Thanks!  Were you able to get a new producer_results.csv and consumer_results.csv after running those commands?  If so, you can just update those in this PR as well

ingomueller-net:
> Thanks! Were you able to get a new producer_results.csv and consumer_results.csv after running those commands? If so, you can just update those in this PR as well

Done! I hadn't understood that those were tracked in the repo but I added them now.

It seems like quite a few results flipped from `True` to `False`. I guess that means that things that work before stopped working? I haven't investigated why that happens but I had looked into quite a few isolated cases before submitting and there didn't seem to be any breakage due to the changes. Still, I am not overall very confident, TBH...

richtia:
> > Thanks! Were you able to get a new producer_results.csv and consumer_results.csv after running those commands? If so, you can just update those in this PR as well
> 
> Done! I hadn't understood that those were tracked in the repo but I added them now.
> 
> It seems like quite a few results flipped from `True` to `False`. I guess that means that things that work before stopped working? I haven't investigated why that happens but I had looked into quite a few isolated cases before submitting and there didn't seem to be any breakage due to the changes. Still, I am not overall very confident, TBH...

Ibis has had a bunch of updates/issues since this was last run, so I can't really speak about those results. As for isthmus...i think I was previously passing all the `skipped` functions that weren't supported in the framework, now they are being skipped altogether.  And skipped tests are marked as failed.  Might be good to look more into this and possibly have a different status (instead of pass/fail) for unsupported functions for each engine.

richtia (approved):

Issue #127, opened 2024-11-05
RFC: decouple pytest/CI failure from failures of systems under test


ingomueller-net:
I would like to propose some changes to the test infrastructure and would like to gather feedback before starting any possible implementation. Please let me know what you think.

## Problem

Since the introduction of the CI tests in #28, the CI tests of this repository have never been green. This defeats their purpose: we cannot rely on them to know whether there are bugs in this repository, which can easily lead to new bugs being introduced without anybody noticing. And exactly that has happened: #123 and #125 have fixed bugs in the test harness that hadn't been noticed before, ~~I have some more in progress,~~ and eyeballing at the failures suggests that there are even more.

## Analysis

I believe that the reason for this is that the current testing logic conflates the failure/pass signal of pytest/CI with the failure/pass signal of the systems under test. However, I argue that it's important to make this distinction: this repository contains non-trivial test logic for Substrait producers and consumers ("systems under test", SUT) and both parts can fail independently. It is paramount that we find failures in the test logic in order to be able to trust it and I argue that **CI (and hence pytest?) should fail if and only if that testing logic breaks**. Failures of systems under test should therefor *not* (always) lead to a failure of CI. If they do, **we can only get CI green if all producers and consumers support the entire Substrait specification**, which we have no realistic hope in achieving ever: they live in external repositories that are not under our control, many of them don't aim at supporting the whole spec, and the various producers and consumers and the spec don't evolve in lock step. I therefor think that failures of systems under test should not lead to CI failure, and we should decouple the two instead.

Visually, this is what is happening currently:

```
             +-------------+                \
pytest ----->| Test case 1 |-----> fail/pass |   ,--- if any tests fails ---> CI fails
    \ \      +-------------+|----> fail/pass  > <
     \ `----->+-------------+|---> fail/pass |   `--- if all tests pass  ---> CI passes
      `------->+-------------+              /
```

## Proposal

In order to get to a state where CI passes as long as the test logic does what we expect (such that we can make CI passing a requirement of each PR), I propose the following changes:

1. **Freeze the dependencies of this repository.** In particular, we need to pull in a frozen version of Substrait such that changes in the Substrait spec don't lead to unexpected failures, and we need to pull in frozen versions of producers, consumers, validators, etc. such that changes in those systems don't lead to unexpected failures (or changes).
[...]

PR #132, opened 2024-11-27
ci: overhaul Github workflows to not use Docker

ingomueller-net:
This PR gets rid of the Docker images and uses Github actions to install the required dependencies instead. This solves several problems:

* The Docker images fixed the versions of some dependencies, so there are now two places where versions are potentially fixd: the Dockerfiles and requriments.txt. The Github actions now also use the latter mechanism, so there is only one. This is something I want to address as per #127.
* The Dockerfile where extremely repetetive: almost all of them installed the same thing and then executing them from the Github repeated again the same thing. The new Github actions uses the `matrix` mechanism to remove the latter source of redundancy and gets rid of the Docker images alltogether.
* (Minor) The Docker images are another source or indirection and somewhat slower. Since they are rebuilt for every CI run, they also don't offer more stability than using Github runners directly.


EpsilonPrime (approved):
Thanks, this should be a lot easier to maintain.

The test failures aren't related to this change.

richtia (approved):

PR #133, opened 2024-11-27
feat: lock versions in requirements.txt

ingomueller-net:
This PR is stacked on and, therefor, includes #132.

This PR locks the versions in `requirements.txt`. The original version of that file is now called `requirements-unlocked` and can be used to update an environment to the latest versions of all packages.

This is one out of two steps towards resolving #127.


EpsilonPrime (commented):
Feels a lot like rust's cargo manifest.  I wonder if dependabot will be able to manage multiple files like this.  But it's worth exploring.

Should we update the documentation to suggest using requirements-unlocked during development?

ingomueller-net:
Yeah, good point. I'll update the doc tomorrow. The workflow I envision is: use the locked file unless the thing you want to do is update a particular version.

EpsilonPrime (approved):

PR #136, opened 2024-11-28
tests: regenerate test snapshots and remove obsolete snapshots

ingomueller-net:
This PR regenerates all tests snapshots. It also removes all snapshots that are currently produced by any tests. I have not investigate whether this is because the tests have been renamed/removed or whether the tests that are supposed to generate these snapshots fail before the get there.


ingomueller-net:
This isn't currently ready to be merge: I regenerated the snapshots with wrong versions of some of the dependencies and need to regenerate them. That turns out to have its own problems, so I'll need a bit to get this PR ready...

substrait_consumer/tests/functional/extension_functions/approximation_snapshots/function_test_results/approx_count_distinct_result.txt
```diff
@@ -1,2 +1,2 @@
-523942
+658483
```
EpsilonPrime, line 1: This seems like something that shouldn't change.

substrait_consumer/tests/functional/relations/set_relation_snapshots/IsthmusProducer/intersect_plan.json
```diff
@@ -131,7 +131,7 @@
               }]
             }
           }],
-          "op": "SET_OP_INTERSECTION_PRIMARY"
+          "op": "SET_OP_INTERSECTION_MULTISET"
```
EpsilonPrime, line 134: Is this right behavior?   (It may be the existing behavior but it may also be wrong.)

EpsilonPrime (approved):
Okay to proceed as is, with the note that some things may need to be marked as wrong later.

substrait_consumer/tests/functional/relations/set_relation_snapshots/IsthmusProducer/intersect_plan.json
ingomueller-net: Good point! I haven't done extensive checks as to whether or not the updates makes sense.

In this specific case, I think that the two are equivalent: "primary" keeps records "that are present in *any* secondary input" and "multiset" keeps records that "that match at least one record from *all* secondary inputs"; since there is only a single secondary input in this query, "any" and "all" is the same.

substrait_consumer/tests/functional/extension_functions/approximation_snapshots/function_test_results/approx_count_distinct_result.txt
ingomueller-net: I would say the opposite: since the function is "approximate," we shouldn't be testing for equality.

It is true, though, that we might require it to be deterministic, so that value shouldn't change *for the same version of the the same consumer* (TBD). But we haven't recorded the version of the consumer until now, so that may very well explain the change...

ingomueller-net:
@EpsilonPrime: We can delay deciding what are the correct results and probably should go through all results again at some point to decide whether they are actually correct but if we can avoid submitting wrong results, I think that'd be good. I thus went again through the diff of the snapshots. I think you covered the most suspicious changes. Here's one more though:

```diff
--- a/substrait_consumer/tests/functional/extension_functions/arithmetic_snapshots/function_test_results/median_result.txt
+++ b/substrait_consumer/tests/functional/extension_functions/arithmetic_snapshots/function_test_results/median_result.txt
@@ -1,2 +1,2 @@
-498.72
+498.73
```

All the remainders seem to be due to (1) the Substrait plan version produced by Isthmus and (2) Isthmus changing the order of `project` and `filter` in the plans it generates.

PR #138, opened 2024-11-28
fix: make DB connection a function-scope fixture

ingomueller-net:
Many tests use a DuckDB connection as a fixture. That fixture was set to a `class` scope, such that a DuckDB connection was created only once for all tests of a particular class, presumably to speed up tests. However, tests may corrupt the DuckDB instance, making all subsequent tests fail as a result. This lack of reproducibility undermines tests and is far worse than slightly longer-running tests. This PR thus makes all mentioned fixtures function-scoped (by removing the `scope` definition, `function` scope is the default).

### DIFF
--- a/substrait_consumer/tests/adhoc/test_adhoc_expression.py
+++ b/substrait_consumer/tests/adhoc/test_adhoc_expression.py
@@ -32,7 +32,7 @@ class TestAdhocExpression:
 
     @staticmethod
     @pytest.fixture(autouse=True)
-    def setup_teardown_class(request):
+    def setup_teardown_function(request):
         cls = request.cls
         cls.produced_plans = set()
 
--- a/substrait_consumer/tests/adhoc/test_adhoc_expression.py
+++ b/substrait_consumer/tests/adhoc/test_adhoc_expression.py
@@ -31,13 +31,13 @@ class TestAdhocExpression:
     """
 
     @staticmethod
-    @pytest.fixture(scope="class", autouse=True)
+    @pytest.fixture(autouse=True)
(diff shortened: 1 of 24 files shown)


substrait_consumer/tests/adhoc/test_adhoc_expression.py
```diff
@@ -31,13 +31,13 @@ class TestAdhocExpression:
     """
 
     @staticmethod
-    @pytest.fixture(scope="class", autouse=True)
+    @pytest.fixture(autouse=True)
     def setup_teardown_class(request):
```
richtia, line 35: Since we're changing the scope to now just be function, let's change the fixture name to be `setup_teardown_function`

richtia: Since there's a lot of overlap in these fixtures, I also wouldn't be opposed to moving it to conftest and having all the function tests reuse the single one.  There would be a little bit a bloat in it though since there are some initializations specific to only some test functions.  This can be handled in a different PR too if you want to do that.

richtia (approved):

substrait_consumer/tests/adhoc/test_adhoc_expression.py
ingomueller-net: Sounds good, I agree with both points: I'll change the function name tomorrow and approve the plan to move the fixture (but will not do it in this PR).

ingomueller-net: Done with latest commit.

PR #139, opened 2024-11-29
feat: decouple conformance failure from pytest failure

ingomueller-net:
This PR implements the main part of #127, the RFC proposing to decouple CI failure from spec adherence. Previously, the tests would record the desired result (Substrait plans, execution results, etc.) in a snapshot using `pytest-snapshot`) and make the test fail if the system under test produced a different result. In this PR, the result of that comparison is now treated as the test "outcome", which, itself, is snapshot and only a change in test *outcome* will now lead to CI failure. This allows to keep CI green at all times even though some systems do not have perfect conformance to the Substrait spec. CI only fails if unexpected things happen: bugs are introduced in the test infrastructure or systems that were previously conformant in a test are now not conformant anymore or vice versa (for example, after upgrading that system to a new version).

Above logic is implemented by wrapping the "inner"/"first" level of snapshots into a function, `check_match`, that, instead of raising an exception if the snapshots don't match, returns a Boolean indicating if the snapshots match. That Boolean is then the outcome of the test, which is snapshot using the normal snapshotting mechanism (which leads to test failure if the snapshots differ).

Also, exceptions produced by producers and consumers due to unsupported plan features are now caught by the test logic and recorded as test "outcome". This way, if a previously unsupported feature is supported by a new version of system under test after an upgrade, the outcome of the test changes from "failure" to "pass" and needs to be changed explicitly by changing the snapshot.

~~The PR depends on and, therefor, currently includes #133, #135, #136, #137, #138, #140, #141, #143, #144, #145, #146, #150, and #151.~~

While this PR probably *is* a step forward but it's also not perfect. I am open as to whether we want to go with the current approach just to get CI green and then think about improving it in subsequent iterations or rather think about how to improve it right away. I *have* spend quite some effort *towards* this PR, but of that has already been merged with #133 to #151, so there is actually only relatively little effort *in* this PR.


ingomueller-net:
Hurray! CI passes for the first time ever 🚀

EpsilonPrime:
Now that all of the dependent PRs have been merged is it time to review?

EpsilonPrime (approved):

PR #145, opened 2024-12-04
ci: add workflow that ensures that support matrix is up-to-date

ingomueller-net:
This PR adds a Github workflow that runs the tests for the support matrix, parses the results, and compares that with the currently checked-in version of the matrix, making CI fail if there is a discrepancy. This ensures that the matrix is always up-to-date with the current test results.

This PR depends on and, therefor, includes #146.


ingomueller-net:
Hold on: one of the targets checking the matrix is failing. Need to fix that...

ingomueller-net:
OK, problem found and fixed in #146. Ready for review :)

EpsilonPrime (approved):

PR #148, opened 2024-12-05
feat: change schema for snapshot paths (WIP)

ingomueller-net:
This PR changes how the paths for snapshots are computed. The main goal is to make it easier to derive what test a certain snapshot belongs to, which I plan to use in #139. The new schema and the way the paths are computed also have a few additional (more arguable?) benefits:

* The path components for the `extension_functions`/`relations` category and the subcategory are used in the `test_*_functions/relations.py` file, where they are available, rather than in `common.py`, where they where previously extracted from the test name string.
* The top-level directory below `extension_functions`/`relations` is simply `snapshots`. Since those are generated files that are rarely modified manually, I argue that having many `*_snapshot` folders were just distracting.
* The next level is now `producer` or `consumer`, which is maybe the single-most distinctive feature, which was previously not represented in the path at all.

The PR also removes a couple of unused function arguments and fixtures.


ingomueller-net:
This hasn't been necessary for the goal I originally wanted to accomplish and is now out of sync...

PR #152, opened 2024-12-09
feat: remove xfail markers for unsupported tests

ingomueller-net:
This PR is based on and, therefor, includes #139.

This PR removes xfail markers that skipped over tests that producers or consumers were thought to not support. With #139, that isn't necessary anymore since there is a different mechanism of dealing with unsupported features.

Some of the features that were presumably unsupported at the time the xfail markers were added have gotten support by the systems under test in the meantime. As a consequence, many of the previously disabled tests now pass. This PR thus also adds snapshots for those files. This change also affects the support matrix -- as far as I can tell, only positively so: I ran the following command and only seem to see flips from `False` to `True`:

```
git diff --word-diff=color --word-diff-regex=",[^,]+" -- app/consumer_results.csv
```

Finally, there are still a few xfail/skip markers left. One is due to a crash in Acero, which was known before; the remaining ones are due to flaky results in the DataFusion consumer, where some tests have (slightly?) different results from run to run such that no snapshot can be found that always works.


EpsilonPrime:
Would merging #139 first making this PR easier to review?

ingomueller-net:
> Would merging #139 first making this PR easier to review?

Yeah, let me merge and rebase.

I find it OK-ish to review stacked PRs by selecting the appropriate range of commits in the "Files Changed" tab. That's also useful for PRs that move a lot of generated files: I have those changes in commits that are separate from the manual changes such that you can look only at the latter.

EpsilonPrime (approved):

PR #154, opened 2024-12-11
fix: use correct arguments for Ibis' `*trim` expressions

ingomueller-net:
This PR fixes the usage of Ibis' `*trim` functions, which do not take any arguments but always remove whitespace instead. Unfortunately, this only changes the error that is produced since `*trim` is currently translated to Substrait correctly (see
ibis-project/ibis-substrait#1224).


EpsilonPrime (approved):
There's a PR in core to add the signature that Ibis uses for trim.  It doesn't really matter here that trim is tested thoroughly (relation testing is more central to consumer-testing) as BFT handles all of the different trim variants.

So this works for me.

PR #156, opened 2024-12-11
feat: store test definitions as data instead of as code

ingomueller-net:
~~This PR is based on and, therefor, includes #155 and #157.~~

This PR moves the test case definitions from the `*_configs.py` files, where they were stored as code, into JSON files such that they are stored as data. The expectation is that this will allow to manipulate test data more conveniently. Also, this paves the way for a subsequent PR that bases test case enumeration solely on the presence of files rather than several levels of indirection in code.

The PR is organized into two separate commits: The first converts the existing `*_configs.py` files into JSON files using a one-off script, which is added by that commit and removed by the second. The second commit then reads the content of the JSON files into the variables in the `*_configs.py` scripts where that data was stored before. Consequently, there is almost no visible change to the outside. Only a small change was necessary in the test case parametrization logic, which relied on the order of the dictionary keys in the test definitions, which has changed during the conversion.


ingomueller-net:
Oh, I still need to update the `*_configs.py` files for the relations tests. Will have to do that tomorrow...

PR #171, opened 2024-12-16
feat: reorganize TPC-H test scripts to make them more similar

ingomueller-net:
~~This PR is based on and, therefor, includes #170.~~

This PR splits the `test_*_tpch.py` scripts into several `test_X_on_Y.py` scripts that have minimal differences among them. More precisely, `test_duckdb_on_acero.py` and `test_duckdb_on_duckdb.py` now only differ on which consumer is being built and `test_duckdb_on_acero.py` and `test_isthmus_on_acero.py` differ only on which plan is used (the former produces it on the fly, the latter loads it from a file). The splitting essentially consisted of moving/copying entire functions from one file to another, changing the order of lines, renaming variables, factoring out common expressions into variables, and changing comments. This is a preparatory step for reusing test functionality in `common.py` in all three scripts.


ingomueller-net:
I forgot to move some snapshots. Will fix and push an updated version.

PR #186, opened 2024-12-18
feat: distinguish plan generation from validation in functional tests

ingomueller-net:
~~This PR is based on and, therefor, includes #184.~~

This PR adds a new test target for the `relation` and `extension_function` producer tests such that now both have one variant that validates the plans whereas the other one doesn't. Validation is useful to see how conformant the producer is. Not activating it is useful to allow non-conformant plans to be generated, which can then be used as input for consumer tests (although it is debatable whether that's a good idea but that's a discussion for another day). The TPC-H tests already do this distinction, and replicating that is the last step before the entry scripts of all categories can be merged.


ingomueller-net:
I had celebrated to early but now CI is actually green 😉

EpsilonPrime (approved):

PR #193, opened 2024-12-20
feat: move and fix adhoc tests

ingomueller-net:
This PR moves the adhoc tests to a more top-level directory, fixes several problems, and adds a target to CI.


.github/workflows/test.yml
```diff
@@ -81,3 +81,7 @@ jobs:
     - name: Run Ibis expression tests
       if: ${{ matrix.producer == 'ibis' }}
       run: pytest substrait_consumer/functional/queries/ibis_expressions/
+
+    - name: Run Ibis expression tests
```
EpsilonPrime, line 85: This is the same name as above -- should it be different?

ingomueller-net: Yes, good catch.

EpsilonPrime (approved):

PR #220, opened 2025-03-27
chore: update `substrait-java` git submodule to latest version

ingomueller-net:
This PR updates the `substrait-java` git submodule to the latest available version, substrait-io/substrait-java@5b12f9b42.


ingomueller-net:
So strange... The problem of not being able to call the Java functions from Python that I observed in substrait-io/substrait-java#351 are now gone and neither of the two commits looks like it explicitly addresses the problem. No idea what has been happening...

EpsilonPrime (approved):

